PR #27, opened 2017-12-09
Typing Annotations

hardbyte:
Python 2 and 3 compatible type annotations for the mypy static type checker.

I've been testing with:
```
pip install mypy
mypy clkhash --ignore-missing-imports --strict-optional --no-implicit-optional --disallow-untyped-calls
```

### DIFF
--- a/clkhash/key_derivation.py
+++ b/clkhash/key_derivation.py
@@ -1,3 +1,5 @@
+from typing import Tuple, List, Union, Optional
+
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.kdf.hkdf import HKDF
 from cryptography.hazmat.backends import default_backend
@@ -71,6 +73,7 @@ class HKDFconfig:
 
     @staticmethod
     def check_is_bytes(value):
+        # type: (Union[str, bytes]) -> bytes
         if isinstance(value, bytes):
             return value
         else:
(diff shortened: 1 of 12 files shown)


clkhash/key_derivation.py
```diff
@@ -71,28 +77,31 @@ def __init__(self, master_secret, salt=None, info=None, hash_algo='SHA256'):
 
     @staticmethod
     def check_is_bytes(value):
+        # type: (Union[str, bytes]) -> bytes
```
wilko77: If you go by the strict definition of this function, then it should be able to accept any type. More like `# type: (Any) -> bytes`

If you look at the context in which we use it in, then your definition is more precise.

However, it kind of seems weird to only allow str and bytes.

clkhash/key_derivation.py
```diff
@@ -71,28 +77,31 @@ def __init__(self, master_secret, salt=None, info=None, hash_algo='SHA256'):
...
         if isinstance(value, bytes):
             return value
         else:
             raise TypeError('provided value is not of type "bytes"')
 
     @staticmethod
     def check_is_bytes_or_none(value):
+        # type: (Optional[bytes]) -> Optional[bytes]
```
wilko77: same here

clkhash/randomnames.py
```diff
@@ -24,12 +26,17 @@
...
+    if data_bytes is None:
+        raise ValueError("No data resource found with name {}".format(resource_name))
+    else:
+        data = data_bytes.decode('utf8')
+        reader = csv.reader(data.splitlines())
+        next(reader, None)  # skip the headers
+        return [row[0] for row in reader]
 
 
 def save_csv(data, schema, file):
```
wilko77: why does this function not have a type annotation?

clkhash/key_derivation.py
hardbyte: Ok will change to `Any`. 

My approach was to try avoid changing code unless it was essential to make the type checker happy. Personally I'd refactor this into `is_bytes` returning a `bool`.

clkhash/randomnames.py
hardbyte: I missed it, and none of the typed functions must call it (otherwise the typechecker would have told me off)

hardbyte: Or maybe I left it as an excercise for the careful reviewer...

hardbyte:
@sjhardy Are you able to review this while @wilko77 is away? I think I've addressed all of his comments.

PR #30, opened 2017-12-13
Add a progress bar for cli based hashing

hardbyte:
Adds a simple progress bar for the command line utility.

screenshot from 2017-12-13 14-35-54

I think the `generate_clk_from_csv` function should expose a better api for progress for other callers though - maybe a progress queue, or callback function?

### DIFF
--- a/clkhash/clk.py
+++ b/clkhash/clk.py
@@ -1,6 +1,8 @@
 """
 Generate CLK from CSV file
 """
+from tqdm import tqdm
+
 import csv
 import logging
 import time
@@ -33,7 +35,7 @@ def hash_and_serialize_chunk(chunk_pii_data, schema_types, keys):
     return clk_data
 
 
-def generate_clk_from_csv(input, keys, schema_types, no_header=False):
+def generate_clk_from_csv(input, keys, schema_types, no_header=False, progress_bar=True):
     log.info("Hashing data")
 
     # Read from CSV file
@@ -53,7 +55,7 @@ def generate_clk_from_csv(input, keys, schema_types, no_header=False):
 
     # Chunks PII
     log.info("Hashing {} entities".format(len(pii_data)))
-    chunk_size = 1000 if len(pii_data) <= 10000 else 10000
+    chunk_size = 200 if len(pii_data) <= 10000 else 1000
 
     # generate two keys for each identifier
     key_lists = generate_key_lists(keys, len(schema_types))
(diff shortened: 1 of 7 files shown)


clkhash/clk.py
```diff
@@ -70,6 +72,22 @@ def generate_clk_from_csv(input, keys, schema_types, no_header=False):
                                          chunk, schema_types, key_lists)
                 futures.append(future)
 
+            if progress_bar:
+                future_iterator = tqdm(concurrent.futures.as_completed(futures),
+                                       desc="Hashing",
+                                       total=len(pii_data) // chunk_size,
+                                       unit='KH',
```
wilko77: this unit is only true if `chunk_size = 1000`

clkhash/clk.py
```diff
@@ -70,6 +72,22 @@ def generate_clk_from_csv(input, keys, schema_types, no_header=False):
                                          chunk, schema_types, key_lists)
                 futures.append(future)
 
+            if progress_bar:
+                future_iterator = tqdm(concurrent.futures.as_completed(futures),
+                                       desc="Hashing",
+                                       total=len(pii_data) // chunk_size,
```
wilko77: you will miss the last chunk if `len(pii_data)` is not a multiple of `chunk_size`

clkhash/clk.py
```diff
@@ -70,6 +72,22 @@ def generate_clk_from_csv(input, keys, schema_types, no_header=False):
...
+                                       leave=False,
+                                       )
+            else:
+                future_iterator = concurrent.futures.as_completed(futures)
+
+            for f in future_iterator:
+                # Note these are not in order of submission - but completion!
+                # We do this noop loop inorder to deal with progress as the futures
+                # complete
+                pass
```
wilko77: This is not very elegant. Now you iterate twice over the futures. And this loop of `pass`.... tststs.... 
Have a loop at my first [stab](https://gist.github.com/wilko77/bba9580ca8aa5b9c98b403002fba7634). I particularly like the `add_done_callback` on futures. 
Also, you don't report progress for the single thread version.

clkhash/clk.py
hardbyte: 😳

clkhash/clk.py
hardbyte: Ahh, I did change the chunk size to 1000 I swear!

Then I forgot why and changed it back...

clkhash/clk.py
hardbyte: I agree your implementation is much better - I've pushed your changes as a commit and then tweaked the callback to be a bit more general.

hardbyte:
Thanks for the code review Wilko, I think I've addresses all your requested changes.

The only piece I'm not happy with in the `generate_clks` function is we should probably split by "executor" and leave it up to the user to decide - and only go back to the if Python version > 3 decision if the user didn't provide a preference. There is of course no reason that Python2 users can't have parallel execution either - I just wanted to use the nicer concurrent api when it exists for Python 3.

Issue #24, opened 2017-12-07
Library documentation


hardbyte:
Generate api docs with [sphinx](http://www.sphinx-doc.org/en/stable/tutorial.html) and expand on notes about clk serialization, use with the linkage service etc.

wilko77:
With PR #36 we have sphinx building the API docs. Yay! However, there are more tasks to do before we can consider _library documentation_ complete. I created issues #37, #38, #39.

PR #35, opened 2018-01-22
recognise the `weight` of an identifier in filter generation

### LINKED ISSUE #18: clk generator does not consider weights
The idea of _weights_ for an identifier is to give it more prevalence in the bloom filter.
The proposed approach of Schnell et.al. is to multiply _k_ by weight. 

_k_ controls the number of different hash functions used to generate the indices in the bloom filter for a given n-gram. Thus, increasing _k_ will result in more bit set to 1 in the bloom filter.

The current implementation ignores the weights.

The identifier format already supports _weights_, so the necessary information is already there, it's just missing the last bit.

wilko77:
weight now adjusts the number of hash functions being used in the Bloom filter generation.

fixes #18

### DIFF
--- a/clkhash/bloomfilter.py
+++ b/clkhash/bloomfilter.py
@@ -33,7 +33,8 @@ def double_hash_encode_ngrams(ngrams,          # type: Iterable[str]
     :param key_md5: hmac secret keys for md5 as bytes
     :param k: number of hash functions to use per element of the ngrams
     :param l: length of the output bitarray
-    :return bitarray of length l with the bits set which correspond to the encoding of the ngrams
+
+    :return: bitarray of length l with the bits set which correspond to the encoding of the ngrams
     """
     bf = bitarray(l)
     bf.setall(False)
@@ -77,10 +78,9 @@ def crypto_bloom_filter(record,         # type: Tuple[Any, ...]
 
     for (entry, tokenizer, key1, key2) in zip(record, tokenizers, keys1, keys2):
         ngrams = [ngram for ngram in tokenizer(entry)]
-        if tokenizer.weight >= 0:
-            adjusted_k = int(round(tokenizer.weight * k))
-        else:
-            adjusted_k = k
+        if tokenizer.weight < 0:
+            raise ValueError('weight must not be smaller than zero, but was: {}'.format(tokenizer.weight))
+        adjusted_k = int(round(tokenizer.weight * k))
         bloomfilter |= double_hash_encode_ngrams(ngrams, key1, key2, adjusted_k, l)
 
     return bloomfilter, record[0], bloomfilter.count()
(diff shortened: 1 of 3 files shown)


clkhash/identifier_types.py
```diff
@@ -15,25 +15,32 @@ class IdentifierType:
...
-        # type: (bool, int, Any) -> None
+        # type: (bool, float, Any) -> None
         """
         :param unigram: Use uni-gram instead of using bi-grams
-        :param int weight: How many times to include this identifier.
+        :param float weight: adjusts the "importance" of this identifier in the Bloom filter.
         :param kwargs: Extra keyword arguments passed to the tokenizer
         Can be set to zero to skip
+
+        Note on weight:
```
hardbyte: This is being a bit pedantic but as eventually (#24) we'd like to have sphinx based docs for clkhash I think this should be formatted with rst:

```
.. note::

  Some note.
```

clkhash/identifier_types.py
```diff
@@ -15,25 +15,32 @@ class IdentifierType:
...
+        For each n-gram of an identifier, we compute k different indices in the Bloom filter which will be set to true.
+        There is a global default_k value, and the k value for each identifier is computed as k = weight * default_k.
+        Reasons why you might want to set weights:
+         - Long identifiers like street name will produce a lot more n-grams than small identifiers like zip code.
+           Thus street name will flip more bits in the Bloom filter and will have a bigger influence in the overall
+           matching score.
+         - The matching might produce better results if identifier that are stable and / or have low error rates are
```
hardbyte: identifier -> identifiers

clkhash/identifier_types.py
```diff
@@ -15,25 +15,32 @@ class IdentifierType:
...
         """
         :param unigram: Use uni-gram instead of using bi-grams
-        :param int weight: How many times to include this identifier.
+        :param float weight: adjusts the "importance" of this identifier in the Bloom filter.
         :param kwargs: Extra keyword arguments passed to the tokenizer
         Can be set to zero to skip
+
+        Note on weight:
+        For each n-gram of an identifier, we compute k different indices in the Bloom filter which will be set to true.
+        There is a global default_k value, and the k value for each identifier is computed as k = weight * default_k.
```
hardbyte: k has to eventually be integer so this should mention if we use `ceil`/`floor`/`round`?

clkhash/bloomfilter.py
```diff
@@ -77,7 +77,11 @@ def crypto_bloom_filter(record,         # type: Tuple[Any, ...]
 
     for (entry, tokenizer, key1, key2) in zip(record, tokenizers, keys1, keys2):
         ngrams = [ngram for ngram in tokenizer(entry)]
-        bloomfilter |= double_hash_encode_ngrams(ngrams, key1, key2, k, l)
+        if tokenizer.weight >= 0:
```
hardbyte: When/Why would the `IdentifierType` weight be negative? Shouldn't we raise an exception or just use k=0 in that case?

PR #36, opened 2018-01-23
Feature docs

wilko77:
can you have a look and tell me if that's sensible? Or am I missing some magic tricks?


docs/conf.py
```diff
@@ -0,0 +1,176 @@
...
+# If true, `todo` and `todoList` produce output, else they produce nothing.
+todo_include_todos = False
+
+
+# -- Options for HTML output ----------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = 'alabaster'
```
hardbyte: I guess we should have the same as [python-paillier](https://github.com/n1analytics/python-paillier/blob/master/docs/conf.py#L30:L35)

docs/conf.py
```diff
@@ -0,0 +1,176 @@
...
+author = 'N1 Analytics'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = '0.8.1'
+# The full version, including alpha/beta/rc tags.
+release = '0.8.1'
```
hardbyte: Could use the version from the library if you want - https://github.com/hardbyte/python-can/blob/develop/doc/conf.py#L18:L25

docs/conf.py
```diff
@@ -0,0 +1,176 @@
...
+# needs_sphinx = '1.0'
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = ['sphinx.ext.autodoc',
+    'sphinx.ext.doctest',
+    'sphinx.ext.intersphinx',
+    'sphinx.ext.coverage',
+    'sphinx.ext.mathjax']
```
hardbyte: If you prefer how the docstrings could be formated (like in python-paillier) we could add [sphinx.ext.napoleon](http://www.sphinx-doc.org/en/stable/ext/napoleon.html). Not too fused - as long as the project is consistent.

PR #40, opened 2018-01-29
Feature blake encoding

wilko77:
this is still w.i.p. but I think it is a good time to get some feedback.

I chose BLAKE2 because it is
- fast
- has a keyed mode, where it allows you to compute a cryptographic MAC with just one hash function call, instead of the two calls in the HMAC construction.

I started on justifying this approach in the docstring, but that still needs some work. 

Timings on my machine with python 3.6.3:
```
comparing n-gram encodings:
double_hash: 12.36605883401353
blake2: 5.57232744700741
```
Yay!


requirements.txt
```diff
@@ -7,3 +7,4 @@ futures==3.1.1
 cryptography==2.1.3
 tqdm==4.19.1
 typing
+pyblake2==1.1.0; python_version < '3.6'
```
hardbyte: Can you add this to the setup.py as well.

Assume the requirements.txt is more for developers but most users will install via `pip install clkhash` which will only follow the instructions in setup.py

wilko77:
can you have another look? Especially at the Enum construction to choose a n-gram encoding... Is that sensible?

Issue #32, opened 2018-01-09
Consider supporting XOR-Folding


hardbyte:
Look into supporting [XOR-Folding for hardening Bloom Filter based Encryptions for PPRL](http://soz-159.uni-duisburg.de/-download=wp-grlc-2016-03.pdf)

nbgl:
Made a new branch for this feature with commit f89aa06.

nbgl:
Implemented in 18e2465. Needs more testing and documentation.

The option is provided as a function argument in the API and a command line argument in the CLI. This should probably be folded into the schema.

nbgl:
This was done in #42.

PR #42, opened 2018-02-06
Feature: XOR folding

nbgl:
As discussed in #32. This implementation adds an extra argument to the CLI, and also an extra argument to the API, permitting the user to specify the number of folds.


nbgl:
I've put the XOR folding in its own function, and changed the tests accordingly. I've also got a small test to check the integration of the XOR folding into existing code.

I would eventually like to include information such as the number of folds, as well as the number of hashes per field and the Bloom filter length, in the schema. It will make more sense to come up with a nice way to structure the hashing settings then. I've opened issue #45 for that.

clkhash/clk.py
```diff
@@ -21,9 +21,10 @@
 log = logging.getLogger('clkhash.clk')
 
 
-def hash_and_serialize_chunk(chunk_pii_data,    # type: Iterable[Tuple[Any]]
-                             schema_types,      # type: Iterable[IdentifierType]
-                             keys               # type: Tuple[Tuple[bytes, ...], Tuple[bytes, ...]]
+def hash_and_serialize_chunk(chunk_pii_data, # type: Iterable[Tuple[Any]]
+                             schema_types,   # type: Iterable[IdentifierType]
+                             keys,           # type: Tuple[Tuple[bytes, ...], Tuple[bytes, ...]]
+                             xor_folds       # type: int
```
wilko77: corresponding docstring needs to be updated with :param xor_folds

nbgl:
XOR folding.

hardbyte:
Can you bump the version to `0.9.0` and update the `CHANGELOG.md` please @nbgl

nbgl:
On it!

PR #54, opened 2018-02-21
Feature: complex schema

nbgl:
This is not quite ready for merging yet, but I would love to hear any and all feedback!


clkhash/bloomfilter.py
```diff
@@ -3,24 +3,51 @@
...
 import hmac
 import sys
-
-from clkhash.identifier_types import IdentifierType
+from typing import Any, Callable, cast, Iterable, List, Sequence, Text, Tuple
 
 from bitarray import bitarray
-from future.builtins import range
-from typing import Tuple, Any, Iterable, List
+from future.builtins import map
```
hardbyte: Might need to explicitly add a dependency on `future` http://python-future.org/quickstart.html#installation

clkhash/bloomfilter.py
```diff
@@ -3,24 +3,51 @@
...
+
+from clkhash import field_formats, tokenizer
+import clkhash.schema
+
+
+try:
+    from_bytes = cast(Callable[[bytes, str], int],  # Close enough.
```
hardbyte: Given this is a testcase for [mypy](https://github.com/python/mypy/blob/master/test-data/unit/check-classes.test#L944) I'm interested why the cast was required?

clkhash/bloomfilter.py
```diff
@@ -3,24 +3,51 @@
...
+        return int(hex_str, 16)
 
 
 def double_hash_encode_ngrams(ngrams,          # type: Iterable[str]
                               key_sha1,        # type: bytes
                               key_md5,         # type: bytes
                               k,               # type: int
-                              l                # type: int
+                              l,               # type: int
+                              encoding         # type: str
```
hardbyte: With function signature updates should be docstring updates.

clkhash/bloomfilter.py
```diff
@@ -40,11 +67,17 @@ def double_hash_encode_ngrams(ngrams,          # type: Iterable[str]
     bf = bitarray(l)
     bf.setall(False)
     for m in ngrams:
-        sha1hm = int(hmac.new(key_sha1, m.encode(), sha1).hexdigest(), 16) % l
-        md5hm = int(hmac.new(key_md5, m.encode(), md5).hexdigest(), 16) % l
+        binary = m.encode(encoding=encoding)
```
hardbyte: I prefer `m_bytes` over `binary`

clkhash/field_formats.py
```diff
@@ -0,0 +1,564 @@
...
+            self.regex = kwargs['regex']
+        if 'case' in kwargs:
+            self.case = kwargs['case']
+        if 'min_length' in kwargs:
+            self.min_length = kwargs['min_length']
+        if 'max_length' in kwargs:
+            self.max_length = kwargs['max_length']
+
+    @classmethod
+    def from_json_dict(self, json_dict):
```
hardbyte: Have a look at https://github.com/ramazanpolat/prodict

clkhash/field_formats.py
```diff
@@ -0,0 +1,564 @@
...
+            raise_from(InvalidEntryError(msg), e)
+
+
+class StringSpec(FieldSpec):
+    """ Represents a field that holds strings.
+
+        One way to specify the format of the entries is to provide a
+        regular expression that they must conform to. Another is to
+        provide zero or more of: minimum length, maximum length, casing
+        (lower, upper, mixed).
```
hardbyte: What about both regex and a min/max length? Why is that not supported?

nbgl: Having both regex and min/max length at the same time is not supported in @wilko77's master schema, so I didn't implement it.

As an aside, it is also possible to use the regex to bound the string length (although this is not practical in all cases as it can result in a much longer regex).

clkhash/schema.py
```diff
@@ -1,36 +1,259 @@
...
+    # type: (Hashable) -> Optional[str]
+    """ Get the path of the master schema given a version.
+
+        :param version: The version of the master schema whose path we
+            wish to retrieve.
+    """
+    try:
+        rel_path = MASTER_SCHEMA_PATHS[version]
+    except (TypeError, KeyError):
+        return None
```
wilko77: why hiding the fact that this call failed when you could throw one of those shiny MasterSchemaErrors?

clkhash/tokenizer.py
```diff
@@ -1,54 +1,44 @@
...
-def bigramlist(word, toremove=None):
-    # type: (str, Optional[List[str]]) -> List[str]
+        :param n: Length of `n`-grams.
+        :param positional: If `True`, then include the index of the
+            substring with the `n`-gram.
+        :raises ValueError: When `n` is negative.
     """
-    Make bigrams from word with pre- and ap-pended spaces
+    if n < 0:
+        raise ValueError('`n` in `n`-gram must be non-negative.')
```
wilko77: what's a 0-gram?

clkhash/tokenizer.py
```diff
@@ -1,54 +1,44 @@
...
-    :param toremove: List of strings to remove before construction
-    :return: list of bigrams as strings
-    """
-    if toremove is not None:
-        for substr in toremove:
-            word = word.replace(substr, "")
-    word = " " + word + " "
-    return [word[i:i+2] for i in range(len(word)-1)]
+    if positional:
+        # Why do these have to be 1-indexed??
```
wilko77: this is the wrong place to ask this question

clkhash/schema.py
```diff
@@ -1,36 +1,259 @@
...
+
+        kdf_salt_b64 = properties_dict['config']['kdf']['salt']
+        result.kdf_salt = base64.b64decode(kdf_salt_b64)
+
+        kdf_info_b64 = properties_dict['config']['kdf']['type']
+        result.kdf_info = base64.b64decode(kdf_info_b64)
+
+        result.xor_folds = 0  # TODO: This will need to change.
+        result.l = 1024  # TODO: This will need to change.
+        result.k = 30  # TODO: This will need to change.
```
wilko77: yes. The defaults should live only in one place. Furthermore, we might want to change the defaults between versions.
A better approach would be to use jsonschema to insert the defaults, if no value given.
See https://python-jsonschema.readthedocs.io/en/latest/faq/#why-doesn-t-my-schema-s-default-property-set-the-default-on-my-instance

requirements.txt
```diff
@@ -2,9 +2,10 @@ bitarray==0.8.1
 nose==1.3.7
 click==6.7
 requests==2.18.4
-PyYAML==3.12
-futures==3.1.1
+futures==3.1.1; python_version < '3.2'
```
wilko77: PyYAML it is needed in some tests.

clkhash/tokenizer.py
nbgl: 0-grams are [permitted in the schema definition](https://github.com/n1analytics/clkhash/blob/master/docs/_static/schema.json#L160), so permitted it here.

clkhash/tokenizer.py
nbgl: Sure. I've modified this comment to be more clear.

clkhash/schema.py
nbgl: I'm wary of validation that modifies the object being validated. I will explore the jsonschema option, but I am sceptical about it.

Would it perhaps be better to have a separate JSON file defining all the defaults that we can refer to from the master schema and from code?

clkhash/schema.py
nbgl: That function (`get_master_schema_path`) deals only with paths. I preferred to leave it to the caller to deal with the exceptional situation of there not being a compatible master schema.

Indeed, in that situation `MasterSchemaError` is thrown in the validation function (`validate_schema_dict`) that calls `get_master_schema_path`.

clkhash/bloomfilter.py
nbgl: Done!

clkhash/bloomfilter.py
nbgl: A cast is needed since `int.from_bytes` has type `Callable[[Arg(Sequence[int], 'bytes'), Arg(str, 'byteorder'), DefaultNamedArg(bool, 'signed')], int]` and there is no syntax for default named arguments in Python 2. (Python 3 has `def foo(bar, *, fizz=buzz`, which forces `fizz` to be named.)

clkhash/bloomfilter.py
nbgl: Indeed, I missed that. Fixed!

clkhash/field_formats.py
nbgl: That would make the code cleaner, but what if I have a '__dict__' key or 'get' key? I wouldn't want to have to keep track of reserved keys when constructing a schema.

clkhash/bloomfilter.py
nbgl: Changed!

clkhash/bloomfilter.py
```diff
@@ -3,24 +3,51 @@
...
+        """
+        if byteorder == 'big':
+            pass
+        elif byteorder == 'little':
+            bytes_ = bytes_[::-1]
+        else:
+            raise ValueError("byteorder must be either 'little' or 'big'")
+
+        hex_str = codecs.encode(bytes_, 'hex')  # type: ignore
+        return int(hex_str, 16)
```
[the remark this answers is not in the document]
nbgl: Done!

clkhash/schema.py
wilko77: returning `None` is considered code smell.
`None` can means many things:
   - Value was never initialised (whether accidentally or on purpose)
   - Value is not valid
   - Value is not needed
   - No such value exists
   - Something went terribly wrong and something that should be there is not
   - …probably dozens of other things

I would find it cleaner to raise an exception here, and let the caller deal with it. 
For instance, in the `validate_schema_dict` function, you cannot give the user a hint of what went wrong: providing `'1'` instead of `1` is different to providing a non-existing version number.

clkhash/schema.py
```diff
@@ -1,36 +1,259 @@
...
+from future.utils import raise_from
+import jsonschema
+
+from clkhash import field_formats
+
+# These are relative to this file. Using tuples to represent hierarches
+# for compatibility (thx Windows for those backslashes).
+MASTER_SCHEMA_PATHS = {1: ('master-schemas', 'v1.json')}  # type: Dict[Hashable, Tuple[Text, ...]]
+
+MASTER_SCHEMA_DIRECTORY = os.path.dirname(__file__)
```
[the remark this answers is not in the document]
wilko77: Since you haven't addressed this, one more question:
Why do you have this tuple with `'master-schemas'` in `MASTER_SCHEMA_PATH`. Do you ever expect to put something different there? If so, when?

clkhash/tokenizer.py
```diff
@@ -1,54 +1,50 @@
...
 
 
-def positional_unigrams(instr):
-    # type: (str) -> List[str]
-    """
-    Make positional unigrams from a word.
 
-    E.g. 1987 -> ["1 1", "2 9", "3 8", "4 7"]
+def get_tokenizer(hash_settings  # type: field_formats.FieldHashingProperties
+                  ):
```
wilko77: needs a docstring

clkhash/schema.py
nbgl: You raise a very good point with the ambiguity. I'll change to raise `MasterSchemaError`.

clkhash/schema.py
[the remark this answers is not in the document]
nbgl: Sorry, I missed this. You are right, it's cleaner for the directory to be `os.path.join(os.path.dirname(__file__), 'master-schemas')`.

tests/test_clk.py
```diff
@@ -19,3 +20,78 @@ def test_uneven_chunk(self):
         self.assertEqual([10, 11, 12, 13, 14, 15, 16], res[1])
 
 
+class TestComplexSchemaChanges(unittest.TestCase):
+    def test_doesnt_crash(self):
+        CSV_INPUT = io.StringIO(
+            'name,id,dob,gender,children\n'
+            'KÉVIN,kev007,1963-12-13,M,1\n'
```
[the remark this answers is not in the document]
nbgl: It doesn't fail because we have `# -*- encoding: utf-8 -*-` and `from __future__ import unicode_literals`.

tests/test_e2e_hashing.py
```diff
@@ -30,16 +35,50 @@ def test_namelist_hashable(self):
...
-        bf0 = bloomfilter.calculate_bloom_filters(pii, it, keys)[0]
-        it = [IdentifierType(weight=1)]
-        bf1 = bloomfilter.calculate_bloom_filters(pii, it, keys)[0]
-        it = [IdentifierType(weight=2)]
-        bf2 = bloomfilter.calculate_bloom_filters(pii, it, keys)[0]
-        it = [IdentifierType(weight=1.5)]
-        bf15 = bloomfilter.calculate_bloom_filters(pii, it, keys)[0]
+
+        schema.fields[0].hashing_properties.weight = 0
+        bf0 = bloomfilter.calculate_bloom_filters(pii, keys, schema)[0]
```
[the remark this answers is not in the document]
nbgl: Thanks, I've fixed that.

tests/test_e2e_hashing.py
```diff
@@ -12,10 +15,12 @@ def test_namelist_hashable(self):
...
 
-        bf1 = bloomfilter.calculate_bloom_filters(s1, namelist.schema_types,
-                                                  generate_key_lists(('secret', 'sshh'), len(namelist.schema_types)))
-        bf2 = bloomfilter.calculate_bloom_filters(s2, namelist.schema_types,
-                                                  generate_key_lists(('secret', 'sshh'), len(namelist.schema_types)))
+        schema = randomnames.NameList.SCHEMA
+        key_lists = generate_key_lists(('secret', 'sshh'),
+                                       len(schema.fields)),
+
+        bf1 = list(bloomfilter.stream_bloom_filters(s1, key_lists, schema))
```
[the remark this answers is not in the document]
nbgl: Yes, that would be cleaner. Done.

clkhash/schema.py
```diff
@@ -1,36 +1,259 @@
...
+
+class GlobalHashingProperties(object):
+    """ Stores global hashing properties.
+
+        :ivar type: TODO: find out what this does.
+        :ivar kdf_type: TODO: find out what this does.
+        :ivar kdf_hash: TODO: find out what this does.
+        :ivar kdf_key_size: TODO: find out what this does.
+        :ivar kdf_salt: TODO: find out what this does.
+        :ivar kdf_info: TODO: find out what this does.
```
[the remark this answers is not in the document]
nbgl: I've updated the documentation so now it documents things.

clkhash/bloomfilter.py
```diff
@@ -95,29 +102,38 @@ def double_hash_encode_ngrams_non_singular(ngrams,          # type: Iterable[str
...
+
+        sha1hm_bytes = hmac.new(key_sha1, m_bytes, sha1).digest()
+        md5hm_bytes = hmac.new(key_md5, m_bytes, md5).digest()
+
+        sha1hm = int_from_bytes(sha1hm_bytes, 'big') % l
+        md5hm = int_from_bytes(md5hm_bytes, 'big') % l
+
         i = 0
         while md5hm == 0:
             md5hm = int(hmac.new(key_md5, m.encode() + chr(i).encode(), md5).hexdigest(), 16) % l
```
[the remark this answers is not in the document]
nbgl: Fixed!

clkhash/cli.py
```diff
@@ -48,12 +48,11 @@ def cli(verbose=False):
...
+@click.argument('schema', type=click.File('r', lazy=True))
 @click.argument('output', type=click.File('w'))
 @click.option('-q', '--quiet', default=False, is_flag=True, help="Quiet any progress messaging")
 @click.option('--no-header', default=False, is_flag=True, help="Don't skip the first row")
-@click.option('--xor-folds', default=0, type=click.IntRange(0, None))
-def hash(input, output, schema, keys, quiet, no_header, xor_folds):
+def hash(input, keys, schema, output, quiet, no_header):
     """Process data to create CLKs
 
     Given a file containing csv data as INPUT, and optionally a json
```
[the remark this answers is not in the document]
nbgl: Thanks, I missed that.

clkhash/tokenizer.py
nbgl: Done!

nbgl:
Moving pull request to #69.

PR #64, opened 2018-03-06
Improved windows support

hardbyte:
Rebases and squashes changes from PR https://github.com/n1analytics/clkhash/pull/63 to better support Windows.

Primarily the changes are:

* concurrent hashing on windows and Python2
* cross platform temporary files for testing purposes.
* testing of CLI on windows

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,3 +30,4 @@ install:
 script:
   - if [ "${MYPY_TYPING}" == "1" ]; then mypy clkhash --ignore-missing-imports --no-implicit-optional --disallow-untyped-calls; fi
   - INCLUDE_CLI= pytest --cov=clkhash
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ matrix:
         - MYPY_TYPING=1
 
 install:
-  - travis_retry pip install pytest pytest-cov
   - if [ "${MYPY_TYPING}" == "1" ]; then travis_retry pip install mypy; fi
   - travis_retry pip install -r requirements.txt
   - travis_retry pip install -e .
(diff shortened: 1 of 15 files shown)


.travis.yml
```diff
@@ -20,13 +20,18 @@ matrix:
       env:
         - TEST_ENTITY_SERVICE=1
         - MYPY_TYPING=1
+        - INCLUDE_CLI=1
+    - python: '2.7'
+      env:
+        - TEST_ENTITY_SERVICE=1
+        - INCLUDE_CLI=1
```
nbgl: Why are we not running Mypy on Python 2?

clkhash/clk.py
```diff
@@ -106,32 +111,25 @@ def generate_clks(pii_data,         # type: Sequence[Tuple[str, ...]]
     # Chunks PII
     log.info("Hashing {} entities".format(len(pii_data)))
     chunk_size = 200 if len(pii_data) <= 10000 else 1000
+    futures = []
```
nbgl: I thought `future`s  had to exist within the context of the executor. It may be better to initialise this list inside the `with` statement.

.travis.yml
hardbyte: Because `mypy` itself requires a newer Python to run - http://mypy.readthedocs.io/en/latest/getting_started.html#installation

clkhash/clk.py
hardbyte: My understanding is that futures can be created with different executors and still function as expected. See for example the [as_completed](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.as_completed) function. But definetely worth calling out to check

PR #72, opened 2018-03-22
Fix generate_subsets method

mhsiah:
The `generate_subsets()` method in `NameList` did not protect against unreasonable parameters. Specifically, if it was asked to generate `n` names in the constructor, then `sz + (1 - overlap)*sz` must be less than `n` for `generate_subsets()` to be able to work. A `ValueError` is now raised if this condition is not met.

This PR also includes a considerable simplification of the code for generating the subsets, as well as a few improvements/clarifications to the documentation.

This PR is a precursor to a solution to entity-service/issues/99.

### DIFF
--- a/clkhash/randomnames.py
+++ b/clkhash/randomnames.py
@@ -137,6 +137,8 @@ class NameList:
 
         :param sz: length of subsets to generate
         :param overlap: fraction of the subsets that should have the same names in them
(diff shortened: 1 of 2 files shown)


hardbyte:
Looks good to me. I'll try deal with the failing appveyor build...

PR #77, opened 2018-03-23
Documentation update

hardbyte:


README.md
```diff
@@ -64,39 +65,3 @@ clks = clk.generate_clk_from_csv(open('fake-pii-out.csv','r'), ('key1', 'key2'),
 $ python -m clkhash.cli benchmark
 10000 hashes in 1.524871 seconds. 6.56 KH/s
 ```
```
mhsiah: The 'cli.rst' file now contains information on running the benchmark. Perhaps it doesn't need to be in both places? If it should be in both places, perhaps the testing info should be too?

docs/cli.rst
```diff
@@ -0,0 +1,81 @@
...
+
+- Using the ``clkutil`` script which should have been added to your path during installation.
+- directly running the python module ``clkhash.cli`` with ``python -m clkhash.cli``.
+
+
+Hashing
+-------
+
+The command line tool ``clkutil`` can be used to hash a csv file of personally identifiable information.
+The tool needs to be provided with keys and a :ref:`schema`; it will output a file containing
```
mhsiah: Minor quibble: 'csv' should probably be capitalised throughout.

docs/cli.rst
```diff
@@ -0,0 +1,81 @@
...
+    0,Libby Slemmer,1933/09/13,F
+    1,Garold Staten,1928/11/23,M
+    2,Yaritza Edman,1972/11/30,F
+
+
+Benchmark
+---------
+
+A quick hashing benchmark can be carried out to determine the rate at which the current machine
+can generate clks from a simple schema::
```
wilko77: Mention that hashing speed does not only depend on the capabilities of your computer, but also on the number of features and the corresponding schema. Thus these numbers will, in general, not be a good predictor for the performance of a specific use-case!

Also, make it clearer which data and schema we use here.

docs/index.rst
```diff
@@ -1,17 +1,25 @@
-.. CLK hash documentation master file, created by
-   sphinx-quickstart on Tue Jan 23 13:22:28 2018.
-   You can adapt this file completely to your liking, but it should at least
-   contain the root `toctree` directive.
 
-Welcome to CLK hash's documentation!
-====================================
+clkhash: Cryptographic Linkage Key Hashing
```
wilko77: doesn't clk stand for cryptographic longterm key?
And I really don't like 'hashing', because the whole process has little to do with hashing.

docs/cli.rst
```diff
@@ -0,0 +1,81 @@
...
+
+    python -m clkhash.cli benchmark
+    generating CLKs: 100%                 10.0K/10.0K [00:01<00:00, 7.72Kclk/s, mean=521, std=34.7]
+     10000 hashes in 1.350489 seconds. 7.40 KH/s
```
wilko77: why does the progress bar show 7.72 KH/s and the summary says 7.40 KH/s?

docs/cli.rst
```diff
@@ -0,0 +1,81 @@
...
+- ``horse staple`` is the secret key that both participants will use to hash their data.
+- ``simple-schema.json`` is a :ref:`schema` describing how to hash the csv. E.g, ignore the first
+  column, use bigram tokens of the name, use positional unigrams of the date of birth etc.
+- ``clk.json`` is the output file.
+
+Data Generation
+---------------
+
+The cli tool has an option for generating fake pii data.
+::
```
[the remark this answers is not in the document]
hardbyte: Sure I avoided because it will change when #69 is merged. I'll just put the 4 line yaml schema definition in the doc.

docs/cli.rst
```diff
@@ -0,0 +1,81 @@
...
+    1,Garold Staten,1928/11/23,M
+    2,Yaritza Edman,1972/11/30,F
+
+It can be hashed using ``clkutil`` with::
+
+    $ clkutil hash --schema simple-schema.json fake-pii.csv horse staple clk.json
+
+Where:
+
+- ``horse staple`` is the secret key that both participants will use to hash their data.
```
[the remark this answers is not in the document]
hardbyte: Sure. I'll capture my thoughts around eventually changing this in https://github.com/n1analytics/clkhash/issues/78

docs/index.rst
hardbyte: I think we have to agree to disagree here.
> A hash function is any function that can be used to map data of arbitrary size to data of fixed size

Issue #67, opened 2018-03-12
`python -m clkhash` not working


nbgl:
It should be enough to rename `cli.py` to `__main__.py`.

hardbyte:
Or we could update the docs to say `python -m clkhash.cli`

nbgl:
As far as I can tell, it is convention that you should be able to run `python -m mymodule` for modules that have a CLI, without having to specify a submodule.

For example, `python -m unittest` gives you the CLI does not specify a submodule. Importing the same module (`import unittest`) doesn’t run the CLI but imports the namespace.

hardbyte:
Good point.

PR #90, opened 2018-04-11
Add a __main__ script so `python -m clkhash` will launch the cli

hardbyte:
Closes issue #67

### DIFF
--- a/clkhash/__main__.py
+++ b/clkhash/__main__.py
@@ -1,0 +1,4 @@
+from .cli import *
+
+if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


clkhash/__main__.py
```diff
@@ -0,0 +1,4 @@
+from .cli import *
```
nbgl: It’s generally bad practice to have star imports in Python. But in such a small file, it’s unlikely to matter.

hardbyte: nah fair point

Issue #88, opened 2018-04-11
validation of the csv header should be optional


wilko77:
Currently, in the `generate_clk_from_csv` function we always check that the csv header names match the corresponding identifier names in the schema.
This should be optional, as, especially in the multiparty case, you cannot assume that every party's csv header aligns with the schema.

nbgl:
Closed by #94.

PR #94, opened 2018-04-12
Option to not check headers against the schema.

nbgl:
This is as per #88.

Also, minor documentation.


clkhash/cli.py
```diff
@@ -51,7 +51,9 @@ def cli(verbose=False):
 @click.argument('output', type=click.File('w'))
 @click.option('-q', '--quiet', default=False, is_flag=True, help="Quiet any progress messaging")
 @click.option('--no-header', default=False, is_flag=True, help="Don't skip the first row")
-def hash(input, keys, schema, output, quiet, no_header):
+@click.option('--no-check-header', default=False, is_flag=True, help="Don't check the header against the schema")
+@click.option('--no-validate', default=False, is_flag=True, help="Don't validate the entries against the schema")
```
wilko77: I guess this is a personal preference thing, but I don't like this double negation here. I would find it more intuitive to change these options to --header, --check-header, etc.

clkhash/clk.py
```diff
@@ -46,35 +47,57 @@ def hash_and_serialize_chunk(chunk_pii_data,  # type: Sequence[Sequence[str]]
...
+            warnings whose aim is to keep the hashes consistent between
+            data sources; this may affect linkage accuracy.
+        :param bool header: Set to `False` if the CSV file does not have
+            a header. Set to `'ignore'` if the CSV file does have a
+            header but it should not be checked against the schema.
+        :param bool progress_bar: Set to `False` to disable the progress
+            bar.
+        :return: A list of serialized Bloom filters and a list of
+            corresponding popcounts.
+    """
```
wilko77: oooOOOoooh. docstring. Me likes it.

clkhash/clk.py
```diff
@@ -46,35 +47,57 @@ def hash_and_serialize_chunk(chunk_pii_data,  # type: Sequence[Sequence[str]]
...
     # Read the lines in CSV file and add it to PII
     pii_data = []
     for line in reader:
-        if len(line) == len(schema.fields):
-            pii_data.append(tuple([element.strip() for element in line]))
-        else:
-            raise ValueError("Line had unexpected number of elements. "
-                "Expected {} but there was {}".format(
-                len(schema.fields), len(line)))
+        pii_data.append(tuple([element.strip() for element in line]))
```
wilko77: we just talked about that. Make sure that we test for the correct amount of element in a line even if validate is set to false.

nbgl: Will do!

PR #114, opened 2018-04-13
This should fix cli.create

nbgl:
`cli.create` used to break because `load_schema` no longer exists. This should fix it.


hardbyte:
Was just about te write this up as an issue:

cli-create

I'll make a release candidate now with this change...

hardbyte:
Now included in https://pypi.org/project/clkhash/0.11.0rc3/

Issue #86, opened 2018-04-11
error messages from validation could be more helpful


wilko77:
example:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/Code/clkhash/clkhash/field_formats.py in validate(self, str_in)
    416         try:
--> 417             value = int(str_in, base=10)
    418         except ValueError as e:

ValueError: invalid literal for int() with base 10: ''

The above exception was the direct cause of the following exception:

InvalidEntryError                         Traceback (most recent call last)
~/Code/clkhash/clkhash/validate_data.py in validate_data(fields, data)
     50             try:
---> 51                 v(entry)
     52             except InvalidEntryError as e:

~/Code/clkhash/clkhash/field_formats.py in validate(self, str_in)
    419             msg = 'Invalid integer. Read {}.'.format(str_in)
--> 420             raise_from(InvalidEntryError(msg), e)
    421 

~/.local/share/virtualenvs/jupyter/lib/python3.6/site-packages/future/utils/__init__.py in raise_from(exc, cause)
    397         execstr = "raise __python_future_raise_from_exc from __python_future_raise_from_cause"
--> 398         exec(execstr, myglobals, mylocals)
    399 

~/Code/clkhash/clkhash/field_formats.py in <module>()

InvalidEntryError: Invalid integer. Read .

The above exception was the direct cause of the following exception:

EntryError                                Traceback (most recent call last)
<ipython-input-19-4565841bd671> in <module>()
      1 from clkhash import clk
----> 2 hashed_data_a = clk.generate_clk_from_csv(a_csv, ('key1', 'key2'), schema)

~/Code/clkhash/clkhash/clk.py in generate_clk_from_csv(input_f, keys, schema, validate, header, progress_bar)
     92                                     keys,
     93                                     validate=validate,
---> 94                                     callback=callback)
     95     else:
     96         results = generate_clks(pii_data,

~/Code/clkhash/clkhash/clk.py in generate_clks(pii_data, schema, keys, validate, callback)
    122 
    123     if validate:
--> 124         validate_data(schema.fields, pii_data)
    125 
    126     # Chunks PII

~/Code/clkhash/clkhash/validate_data.py in validate_data(fields, data)
     51                 v(entry)
     52             except InvalidEntryError as e:
---> 53                 raise_from(EntryError('Invalid entry.'), e)
     54 
     55 

[...]

PR #116, opened 2018-04-16
Feature useful schema error messages

nbgl:
As requested in #86.


clkhash/cli.py
```diff
@@ -76,14 +76,20 @@ def hash(input, keys, schema, output, quiet, no_header, check_header, validate):
...
+    try:
+        clk_data = clk.generate_clk_from_csv(
+            input, keys, schema_object,
+            validate=validate,
+            header=header,
+            progress_bar=not quiet)
+    except (validate_data.EntryError, validate_data.FormatError) as e:
+        msg, = e.args
+        print(msg)
+        print('Hashing failed.')
```
wilko77: these should be `log` not `print`

PR #118, opened 2018-04-17
Feature update cli docs

wilko77:
- split the tutorial notebook into two, one using the Python API, one with the CLI. We do the full matching procedure now, with computing precision and recall at the end.
- Minor changes to the CLI docs to mirror changes to schema.

The notebook should be re-run before merging, as both, anonlink and entity service contained bugs at the time of the run which affect the matching!


hardbyte:
👍

PR #120, opened 2018-04-17
Fix integer validation.

### LINKED ISSUE #119: IntegerSpec validation permits numbers surrounded by some non-numerics
IntegerSpec permits the following strings: `    42`, `42    `, `+42`.

Whilst these are non-ambiguous integers, we don’t want to tokenise and hash whitespace and superfluous pluses. Thus, these should not be permitted.

nbgl:
No longer permit strings such as ' 42', '42 ', or '+42'. Closes #119.

### DIFF
--- a/clkhash/field_formats.py
+++ b/clkhash/field_formats.py
@@ -10,6 +10,7 @@ from __future__ import unicode_literals
 import abc
 from datetime import datetime
 import re
+import string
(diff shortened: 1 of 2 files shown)


hardbyte:
Will we handle negative numbers?

nbgl:
Negative numbers are not permitted by the master schema, so that’s out of the scope of this issue.

PR #122, opened 2018-04-17
Feature support missing values

wilko77:
PII is not always perfect. It can contain missing values. These are either benign blanks or some malicious things like 'N/A', 'null', ... .
In any case, with our fancy validation clkhash falls over if the data contains missing values.

This PR introduces a feature to enable users to define allowed missing values for each field.
And as some of those missing values are clearly bonkers, one can optionally replace them with a more sane value.

All of this, nicely packaged from schema to docs to tests for you to enjoy.
Be amazed!


clkhash/field_formats.py
```diff
@@ -33,6 +33,32 @@ class InvalidSchemaError(ValueError):
...
+        # type: (...) -> None
+        self.sentinel = sentinel
+        self.replace_with = replace_with if replace_with is not None else sentinel
+
+    @classmethod
+    def from_json_dict(cls, json_dict):
+        # type: (Dict[str, Any]) -> MissingValueSpec
+        return cls(
+            sentinel=json_dict['sentinel'],
+            replace_with=json_dict.get('replaceWith', None)
```
nbgl: The `json_dict.get('replaceWith')` is enough—the `, None` is redundant.

PR #126, opened 2018-04-19
Use iterators to avoid unnecessary memory allocation

nbgl:
In a bunch of places we’ve been unnecessarily allocating memory by making lists. This pull request does the following two things to avoid that:

- Imports `range`, `map`, `zip` from `future` when appropriate. These are the versions that are iterators.
- Recognises that `tuple` and `set` will happily accept an iterator, so no need to make a list first.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -10,6 +10,7 @@ import time
 import unittest
 
 from click.testing import CliRunner
+from future.builtins import range
 
 import clkhash
 import clkhash.cli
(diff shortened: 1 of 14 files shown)


tests/test_cli.py
```diff
@@ -108,7 +109,7 @@ class BasicCLITests(unittest.TestCase):
     def test_list_commands(self):
         runner = CliRunner()
         result = runner.invoke(clkhash.cli.cli, [])
-        for expected_command in set(['hash', 'upload', 'create', 'results', 'generate', 'benchmark']):
+        for expected_command in 'hash', 'upload', 'create', 'results', 'generate', 'benchmark':
```
hardbyte: I haven't seen that before

nbgl:
PyPy3 had crashed again, so I restarted the job. It should be happy again soon.

tests/test_cli.py
nbgl: Yeah, it really looks like you would need brackets around that tuple.

PR #128, opened 2018-04-20
support for negative integers

wilko77:
two things:
- we now support negative integers. (You can always look at the positive side of things...)
- we are more forgiving on the input integer strings. The problem was that there are several valid integer strings for on integer. E.g.: `' 1'`, `'+1'`, `'1 '` are all valid and parse-able strings for the integer `1`. We now first parse the integer string to an integer, then pass it on for hashing without white-spaces and `+`.

### DIFF
--- a/clkhash/field_formats.py
+++ b/clkhash/field_formats.py
@@ -557,7 +557,7 @@ class IntegerSpec(FieldSpec):
         """
         try:
             value = int(str_in, base=10)
-            return '{}'.format(value)
+            return str(value)
         except ValueError as e:
             msg = "Invalid integer. Read '{}'.".format(str_in)
             e_new = InvalidEntryError(msg)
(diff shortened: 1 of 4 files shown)


clkhash/field_formats.py
```diff
@@ -553,6 +545,25 @@ def validate(self, str_in):
...
+            strategy of unification here is to first parse the integer string to an Integer type. Thus all of
+            '+13', ' 13', '13' will be parsed to 13. We then convert the integer value to an unambiguous string
+            (no whitespaces, leading '-' for negative numbers, no leading '+').
+
+            :param str_in: integer string
+            :return: integer string without whitespaces, leading '-' for negative numbers, no leading '+'
+        """
+        try:
+            value = int(str_in, base=10)
+            return '{}'.format(value)
```
nbgl: I think str(value) would be simpler.

Issue #99, opened 2018-04-13
Code formatting issues - automate?


hardbyte:
Consider applying [black](https://github.com/ambv/black)

hardbyte:
Black have [patched](https://github.com/python/black/pull/642/files) to now handle type annotations so could we worth trying this again.

PR #142, opened 2018-05-04
Automated code format update

hardbyte:
Does the manual component of #99.

I just selected all files and asked pycharm to clean em up


clkhash/bloomfilter.py
```diff
@@ -30,11 +30,11 @@
...
-def double_hash_encode_ngrams(ngrams,          # type: Iterable[str]
-                              keys,            # type: Sequence[bytes]
-                              k,               # type: int
-                              l,               # type: int
-                              encoding         # type: str
+def double_hash_encode_ngrams(ngrams,  # type: Iterable[str]
+                              keys,  # type: Sequence[bytes]
+                              k,  # type: int
+                              l,  # type: int
+                              encoding  # type: str
```
nbgl: In Mypy’s [Python 2 examples](http://mypy.readthedocs.io/en/latest/python2.html), the type annotations are aligned like in the original.

clkhash/bloomfilter.py
```diff
@@ -5,21 +5,21 @@
...
 import base64
-from enum import Enum
-from functools import partial
-from hashlib import md5, sha1
 import hmac
 import math
 import struct
+from enum import Enum
+from functools import partial
+from hashlib import md5, sha1
```
nbgl: Seems that it’s listing namespace imports (`import foo`) before from imports (`from foo import bar`). This behaviour doesn’t seem to be part of PEP8. Looks better though!

clkhash/schema.py
```diff
@@ -188,15 +188,14 @@ def from_json_file(cls, schema_file, validate=True):
         try:
             schema_dict = json.load(schema_file)
         except ValueError as e:  # In Python 3 we can be more specific
-                                 # with json.decoder.JSONDecodeError,
-                                 # but that doesn't exist in Python 2.
+            # with json.decoder.JSONDecodeError,
+            # but that doesn't exist in Python 2.
```
nbgl: Yeah, fair enough. This comment should be on its own line in the first place.

clkhash/schema.py
```diff
@@ -217,14 +216,14 @@ def get_master_schema(version):
 
     try:
         schema_bytes = pkgutil.get_data('clkhash',
-            'master-schemas/{}'.format(file_name))
+                                        'master-schemas/{}'.format(file_name))
```
nbgl: Good!

clkhash/bloomfilter.py
hardbyte: I had a quick look and it didn't appear to be configurable 👎 

I'm not particularly keen to go through and manually edit... if you feel strongly about it do you want to make any changes (or look into other tools that might reformat more optimally)

hardbyte:
As this got left behind and we had another PR which addressed code formating I've force pushed. Would you mind taking another quick look @nbgl

PR #145, opened 2018-05-08
Remove HKDFConfig

### LINKED ISSUE #136: Get rid of HKDFconfig
`HKDFconfig` not used anywhere except tests of itself. Its functionality is duplicated by `schema.GlobalHashingProperties`.

nbgl:
`HKDFConfig` was not being used anywhere except in key_derivation.py. Even there, it was only used as a struct to pass arguments from one function to another.

The key derivation API will be much cleaner with it gone. I’ve moved the documentation of the parameters to the `hkdf` function. I’ve also removed some tests that are no longer necessary.

Closes #136.

### DIFF
--- a/clkhash/key_derivation.py
+++ b/clkhash/key_derivation.py
@@ -11,111 +11,67 @@ We use the block-size of SHA1 and MD5 as the default key size for HMAC
 """
 DEFAULT_KEY_SIZE = 64
 
+_HASH_FUNCTIONS = {
+    'SHA256': hashes.SHA256,
+    'SHA512': hashes.SHA512
+}
 
-class HKDFconfig:
-    supported_hash_algos = 'SHA256', 'SHA512'
-
-    def __init__(self,
-                 master_secret,   # type: bytes
-                 salt=None,             # type: Optional[bytes]
-                 info=None,             # type: Optional[bytes]
-                 hash_algo='SHA256'     # type: str
-                 ):                     # type: (...) -> None
-        """
-        The parameters for the HDKF are defined as follows:
-
-        master_secret
-            input keying material
-
-        salt
-            HKDF is defined to operate with and without random salt.  This is
-            done to accommodate applications where a salt value is not available.
-            We stress, however, that the use of salt adds significantly to the
-            strength of HKDF, ensuring independence between different uses of the
-            hash function, supporting "source-independent" extraction, and
-            strengthening the analytical results that back the HKDF design.
-
-            Random salt differs fundamentally from the initial keying material in
-            two ways: it is non-secret and can be re-used.
-
-            Ideally, the salt value is a random (or pseudorandom) string of the
-            length HashLen.  Yet, even a salt value of less quality (shorter in
-            size or with limited entropy) may still make a significant
-            contribution to the security of the output keying material.
-
-        info
-            While the 'info' value is optional in the definition of HKDF, it is
-            often of great importance in applications.  Its main objective is to
-            bind the derived key material to application- and context-specific
-            information.  For example, 'info' may contain a protocol number,
-            algorithm identifiers, user identities, etc.  In particular, it may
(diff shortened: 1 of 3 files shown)


clkhash/key_derivation.py
```diff
@@ -11,111 +11,67 @@
...
-        'SHA512': hashes.SHA512
-    }
-    if not isinstance(hkdf_config, HKDFconfig):
-        raise TypeError('provided config has to be of type "HKDFconfig"')
-    hkdf = HKDF(algorithm=hash_dict[hkdf_config.hash_algo](), length=num_keys * key_size, salt=hkdf_config.salt,
-                info=hkdf_config.info, backend=default_backend())
-    # hkdf.derive returns a block of num_keys * key_size bytes which we divide up into num_keys chunks,
-    # each of size key_size
-    keybytes = hkdf.derive(hkdf_config.master_secret)
+    hkdf = HKDF(algorithm=_HASH_FUNCTIONS[hash_algo](),
```
wilko77: if you provide an unsupported hash function, you'll get a not so informative error message:
```
>           hkdf('master_secret'.encode(), 12, key_size=64, hash_algo='MD5')

test_key_derivation.py:134: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def hkdf(master_secret,             # type: bytes
             num_keys,                  # type: int
             hash_algo='SHA256',        # type: str
             salt=None,                 # type: Optional[bytes]
             info=None,                 # type: Optional[bytes]
             key_size=DEFAULT_KEY_SIZE  # type: int
             ):
        # type: (...) -> Tuple[bytes, ...]
        """
        Executes the HKDF key derivation function as described in rfc5869 to
        derive `num_keys` keys of size `key_size` from the master_secret.
    
        :param master_secret: input keying material
        :param num_keys: the number of keys the kdf should produce
        :param hash_algo: The hash function used by HKDF for the internal
            HMAC calls. The choice of hash function defines the maximum
            length of the output key material. Output bytes <= 255 * hash
            digest size (in bytes).
        :param salt: HKDF is defined to operate with and without random
            salt. This is done to accommodate applications where a salt
            value is not available. We stress, however, that the use of salt
            adds significantly to themstrength of HKDF, ensuring
            independence between different uses of the hash function,
            supporting "source-independent" extraction, and strengthening
            the analytical results that back the HKDF design.
                Random salt differs fundamentally from the initial keying
            material in two ways: it is non-secret and can be re-used.
                Ideally, the salt value is a random (or pseudorandom) string
            of the length HashLen.  Yet, even a salt value of less quality
            (shorter in size or with limited entropy) may still make a
            significant contribution to the security of the output keying
            material.
        :param info: While the 'info' value is optional in the definition of
            HKDF, it is often of great importance in applications. Its main
            objective is to bind the derived key material to application-
            and context-specific information. For example, 'info' may
            contain a protocol number, algorithm identifiers, user
            identities, etc.  In particular, it may prevent the derivation
            of the same keying material for different contexts (when the
            same input key material (IKM) is used in such different
            contexts).  It may also accommodate additional inputs to the key
            expansion part, if so desired (e.g., an application may want to
            bind the key material to its length L, thus making L part of the
            'info' field).  There is one technical requirement from 'info':
            it should be independent of the input key material value IKM.
        :param key_size: the size of the produced keys
        :return: Derived keys
        """
>       hkdf = HKDF(algorithm=_HASH_FUNCTIONS[hash_algo](),
                    length=num_keys * key_size,
                    salt=salt,
                    info=info,
                    backend=default_backend())
E       KeyError: 'MD5'

../clkhash/key_derivation.py:67: KeyError
```
I prefer to have a more helpful error message, like we had before.

nbgl: Good point! I’ve updated this with a nice error message.

PR #147, opened 2018-05-10
Code formatting changes following internal code review

hardbyte:
Changes from Brendon after running an internal code review.

### DIFF
--- a/clkhash/field_formats.py
+++ b/clkhash/field_formats.py
@@ -10,7 +10,6 @@ from __future__ import unicode_literals
 import abc
 from datetime import datetime
 import re
-import string
 from typing import Any, cast, Dict, Iterable, Optional, Text, Union
 
 from future.builtins import range, super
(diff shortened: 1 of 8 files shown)


clkhash/field_formats.py
```diff
@@ -325,23 +325,25 @@ def __init__(self,
...
         if regex_based:
             regex_str = cast(str, regex)
             try:
                 compiled_regex = re_compile_full(regex_str)
+                self.regex = compiled_regex
             except (SyntaxError, re.error) as e:
                 msg = "invalid regular expression '{}.'".format(regex_str)
                 e_new = InvalidEntryError(msg)
                 e_new.field_spec = self
                 raise_from(e_new, e)
```
nbgl: Why this change? Does PyCharm not understand `raise_from`?

wilko77:
I pushed some additional changes. 

Some of the proposed `noinspection` annotations do not seem to be needed any more. However, I found a few more issues where those are needed to pass the compatibility check.

So in total it is a net loss in terms of `# noinspection` pollution.

To make up for it I fixed some PIP violations and out-of-date docstrings.

PR #148, opened 2018-05-24
Feature update command line tool to use new Entity Service REST api

hardbyte:


clkhash/cli.py
```diff
@@ -221,19 +260,19 @@ def upload(input, mapping, apikey, server, output, verbose):
         log(response.text)
         log("When the other party has uploaded their CLKS, you should be able to watch for results")
```
wilko77: this log message is not quite right any more.

clkhash/cli.py
```diff
@@ -157,39 +154,81 @@ def create(type, schema, server, output, threshold, verbose):
...
-        log("Mapping created")
-        if verbose:
-            log(MAPPING_CREATED_MSG.format(**response.json()))
         print(response.text, file=output)
 
 
 @cli.command('upload', short_help='upload hashes to entity service')
 @click.argument('input', type=click.File('r'))
-@click.option('--mapping', help='Server identifier of the mapping')
+@click.option('--project', help='Server identifier of the mapping')
```
wilko77: no mapping. it's a project. Most commands have a `--project` option, why do those options all have different help strings?

clkhash/cli.py
```diff
@@ -250,7 +289,7 @@ def results(mapping, apikey, watch, server, output):
 
     def get_result():
         return requests.get(
-            '{}/api/v1/mappings/{}'.format(server, mapping),
+            '{}/api/v1/projects/{}/runs/{}/result'.format(server, project, run),
             headers={"Authorization": apikey}
         )
```
wilko77: the following while loop might run forever, as a run result can have other terminal states than 200. 
Better: `while response.status_code == 404:`

tests/test_cli.py
```diff
@@ -372,120 +372,141 @@ def tearDown(self):
...
 
-        self.assertGreaterEqual(len(out['resource_id']), 16)
+        self.assertGreaterEqual(len(out['project_id']), 16)
         self.assertGreaterEqual(len(out['result_token']), 16)
         self.assertGreaterEqual(len(out['update_tokens']), 2)
 
+    def test_create_with_bad_schema(self):
         # Make sure we don't succeed with bad schema.
         runner = CliRunner()
         with temporary_file() as output_filename:
```
wilko77: what do you need `output_filename` for?

clkhash/cli.py
```diff
@@ -157,39 +154,81 @@ def create(type, schema, server, output, threshold, verbose):
         schema_json = json.load(schema)
         clkhash.schema.Schema.from_json_dict(schema_json)
```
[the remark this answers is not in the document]
hardbyte: Fair point, it is essentially to validate.

I'll replace with `validate_schema_dict`

clkhash/cli.py
```diff
@@ -157,39 +154,81 @@ def create(type, schema, server, output, threshold, verbose):
...
     else:
-        schema_json = 'NOT PROVIDED'
+        raise ValueError("Schema must be provided when creating new linkage project")
```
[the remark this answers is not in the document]
hardbyte: Well in the entity service OpenAPI spec we **require** a schema to be given when creating a new project.

The cli might be able to pass a bogus or default schema (e.g. `{}`) but I thought the server now validates it too. I'll leave it as strict for now.

tests/test_cli.py
hardbyte: Ahh good catch. I'll inspect the output...

clkhash/cli.py
nbgl: This looks done.

clkhash/cli.py
```diff
@@ -157,39 +154,81 @@ def create(type, schema, server, output, threshold, verbose):
...
 @cli.command('upload', short_help='upload hashes to entity service')
 @click.argument('input', type=click.File('r'))
-@click.option('--mapping', help='Server identifier of the mapping')
+@click.option('--project', help='Server identifier of the mapping')
 @click.option('--apikey', help='Authentication API key for the server.')
 @click.option('--server', type=str, default=DEFAULT_SERVICE_URL, help="Server address including protocol")
 @click.option('-o','--output', type=click.File('w'), default='-')
 @click.option('-v', '--verbose', default=False, is_flag=True, help="Script is more talkative")
-def upload(input, mapping, apikey, server, output, verbose):
+def upload(input, project, apikey, server, output, verbose):
```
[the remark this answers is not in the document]
nbgl: This also looks done.

PR #159, opened 2018-06-30
Update cli and abstract rest calls into own module

hardbyte:

PR #161, opened 2018-07-04
Feature test rest client

hardbyte:
I felt guilty about not fully testing the rest_client abstraction in #159 

Also this testing seems to have exposed a bug in the entity service :-)


hardbyte:
This is failing CI tests because it has identified a bug in the entity service. Will merge here and fix upstream

PR #162, opened 2018-07-04
Create two helper functions to wait for a run to complete

hardbyte:
Extra feature for the rest_client - wait for run completion.

One function yields status updates, one just waits until the very end (or until a timeout expires).

~I could change the generator to only give updates that have "changed" - not just every second. But interested in your thoughts.~  Now implemented.

### DIFF
--- a/clkhash/backports.py
+++ b/clkhash/backports.py
@@ -104,8 +104,12 @@ unicode_reader = (_p2_unicode_reader  # Python 2 with hacky workarounds.
                   else csv.reader)  # Py3 with native Unicode support.
 
 if sys.version_info > (3, 2):
-    strftime = datetime.strftime
+    TimeoutError = globals()['__builtins__']['TimeoutError']
+else:
+    TimeoutError = Exception
 
+if sys.version_info > (3, 2):
+    strftime = datetime.strftime
 else:
     _YEAR_LEN = 4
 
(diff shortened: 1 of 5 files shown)


clkhash/backports.py
```diff
@@ -104,8 +104,12 @@ def _p2_unicode_reader(unicode_csv_data, dialect=csv.excel, **kwargs):
                   else csv.reader)  # Py3 with native Unicode support.
 
 if sys.version_info > (3, 2):
-    strftime = datetime.strftime
+    TimeoutError = globals()['__builtins__']['TimeoutError']
+else:
+    TimeoutError = Exception
```
nbgl: If someone wants to catch this then `try TimeoutError` will catch everything. Use `OSError` instead of `Exception`.

clkhash/backports.py
```diff
@@ -104,8 +104,12 @@ def _p2_unicode_reader(unicode_csv_data, dialect=csv.excel, **kwargs):
                   else csv.reader)  # Py3 with native Unicode support.
 
 if sys.version_info > (3, 2):
-    strftime = datetime.strftime
+    TimeoutError = globals()['__builtins__']['TimeoutError']
```
nbgl: Suggested alternative:
```python
try:
    TimeoutError = __builtins__.TimeoutError
except AttributeError:
    # Deal with the backport
```

hardbyte: Yeah I started with something like this but mypy was unhappy.

```
mypy clkhash --ignore-missing-imports --strict-optional --no-implicit-optional --disallow-untyped-calls
clkhash\backports.py:109: error: Cannot assign to a type
clkhash\backports.py:109: error: Cannot assign multiple types to name "TimeoutError" without an explicit "Type[...]" annotation
clkhash\backports.py:109: error: Incompatible types in assignment (expression has type "Type[OSError]", variable has type "Type[TimeoutError]")
```

Issue #176, opened 2018-08-09
Support binary uploads to entity-service


hardbyte:
An experimental api has been added for uploading CLKs as a binary file. This is to allow for faster and more efficient data transfer. The same rest endpoint (`/projects/{project_id}/clks`) is used.

Additional headers are required:

    Content-Type: application/octet-stream
    Hash-Count: XX
    Hash-Size: YY

The uploaded stream is simply a packed serialization of (index, CLK data, popcount):

```python
bit_packing_fmt = "!1I<Hash-Size>s1H"
```

Where:

-  `"!"` Use network byte order (big-endian).
- `"1I"` Store the index in 4 bytes as an unsigned int.
- `<Hash-Size>s` the length of a hash in Bytes. Default value is "128s". Store the 128 raw bytes of the bitarray.
-  `"1H"` The popcount stored in 2 bytes (unsigned short)

https://docs.python.org/3/library/struct.html#format-strings

Aha! Link: https://csiro.aha.io/features/ANONLINK-26

hardbyte:
Clkhash library isn't coupled to the entity-service anymore

PR #187, opened 2018-09-04
Cli describe

neilbacon:
Questions in # TODO: comments.

Works: python{,3} -m clkhash.cli describe fake-clk.json

Works: python3 -m unittest tests/test_bloomfilter.py
but with python (2):
  ...
  File "/usr/lib/python2.7/unittest/loader.py", line 91, in loadTestsFromName
    module = __import__('.'.join(parts_copy))
ImportError: Import by filename is not supported.

Type checker not yet run.

All improvements/comments welcome.

### DIFF
--- a/clkhash/bloomfilter.py
+++ b/clkhash/bloomfilter.py
@@ -352,3 +352,13 @@ def serialize_bitarray(ba):
 
     """
     return base64.b64encode(ba.tobytes()).decode('utf8')
+
+# TODO: I think this must already exist in anonlink - maybe that implelemtation should be moved here
+def deserialize_bitarray(ser):
+    # type: str -> (bitarray)
+    """Deserialize a base 64 encoded string to a bitarray (bloomfilter)
+    
+    """
+    ba = bitarray()   # TODO: Is endianess used consistently?
+    ba.frombytes(base64.b64decode(ser.encode(encoding='UTF-8', errors='strict')))
+    return ba
--- a/clkhash/bloomfilter.py
+++ b/clkhash/bloomfilter.py
@@ -355,7 +355,7 @@ def serialize_bitarray(ba):
 
 # TODO: I think this must already exist in anonlink - maybe that implelemtation should be moved here
 def deserialize_bitarray(ser):
-    # type: str -> (bitarray)
+    # type: (str) -> bitarray
     """Deserialize a base 64 encoded string to a bitarray (bloomfilter)
     
     """
--- a/clkhash/bloomfilter.py
+++ b/clkhash/bloomfilter.py
@@ -345,20 +345,3 @@ def stream_bloom_filters(dataset,  # type: Iterable[Sequence[Text]]
                                 keys, hash_properties)
             for s in dataset)
 
-
-def serialize_bitarray(ba):
-    # type: (bitarray) -> str
-    """Serialize a bitarray (bloomfilter)
-
-    """
-    return base64.b64encode(ba.tobytes()).decode('utf8')
-
-# TODO: I think this must already exist in anonlink - maybe that implelemtation should be moved here
-def deserialize_bitarray(ser):
-    # type: (str) -> bitarray
-    """Deserialize a base 64 encoded string to a bitarray (bloomfilter)
-    
-    """
-    ba = bitarray()   # TODO: Is endianess used consistently?
-    ba.frombytes(base64.b64decode(ser.encode(encoding='UTF-8', errors='strict')))
-    return ba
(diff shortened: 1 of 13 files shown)


clkhash/bloomfilter.py
```diff
@@ -352,3 +352,13 @@ def serialize_bitarray(ba):
 
     """
     return base64.b64encode(ba.tobytes()).decode('utf8')
+
+# TODO: I think this must already exist in anonlink - maybe that implelemtation should be moved here
```
hardbyte: Good point - probably into a `serialization.py` module though

setup.py
```diff
@@ -13,7 +13,8 @@
         "jsonschema>=2.6",
         "requests>=2.18",
         "tqdm>=4.24",
-        "typing>=3.6; python_version < '3.5'"  # Backport from Py3.5
+        "typing>=3.6; python_version < '3.5'",  # Backport from Py3.5
+	"bashplotlib>=0.6.5"
```
hardbyte: Think a tab snuck in here

clkhash/describe.py
```diff
@@ -0,0 +1,27 @@
...
+import json
+from clkhash.backports import raise_from
+from bashplotlib.histogram import plot_hist
+from clkhash.bloomfilter import deserialize_bitarray
+
+class DescribeError(Exception):
+    """ The user provided CLK JSON is invalid.
+    """
+
+def plot(input):
```
hardbyte: I like this - much cleaner having the command implemented in its own module.

clkhash/cli.py
```diff
@@ -270,6 +270,12 @@ def benchmark():
     bench.compute_hash_speed(10000)
 
 
+@cli.command('describe', short_help='show distribution of clk popcounts')
+@click.argument('input', type=click.File('r'))
+def describe(input):
```
hardbyte: Don't use `input` as it shadows the builtin `input` function.

Consider adding a `--no-plot` flag

neilbacon: Changed input to clk_json in describe.py.

In cli.py `input` is also used in `hash` and `upload` commands (I copied). Changing it requires a matching change to `@click.argument('input', type=click.File('r'))` which changes the `cli cmd --help` output. Do we want to do that? If so how about:
hash: rename input -> pii_csv
upload: rename input -> clk_json
describe: rename input -> clk_json

hardbyte: Sure let's not bother with reimplementing the summary table.

Your proposed parameter names sound good ' up to you if you'd like to make the change for the other commands.

clkhash/describe.py
hardbyte: Yeah I can't tell what the trailing digit is, I guess the numbers are better than nothing!

Regarding testing I'd just like a simple smoketest - something which uses the output file from the hash command and tries to `describe` it.

clkhash/bloomfilter.py
neilbacon: OK I'll move serialization to serialization.py.

I've noticed inconsistent presence of: 
#!/usr/bin/env python3
or #!/usr/bin/env python3.6
or nothing at top of files.
Should that be removed?

clkhash/cli.py
hardbyte: Yikes! I'm glad you're catching all of these. Yes those suggestions all sound fine. 

Unfortunately some of these outputs *could* change soon with the addition of supporting a binary clk file. #176

clkhash/bloomfilter.py
hardbyte: Sure, I don't feel strongly at all about those

clkhash/describe.py
neilbacon: OK, test added, discovered StringIO and setting sys.stdout in the process.

clkhash/cli.py
neilbacon: these changes now in cli.py (added pii_schema.json to example cmd line too)

clkhash/serialization.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def serialize_bitarray(ba):
+    # type: (bitarray) -> str
+    """Serialize a bitarray (bloomfilter)
+
+    """
+    return base64.b64encode(ba.tobytes()).decode('utf8')
+
+# TODO: I think this must already exist in anonlink - maybe that implelemtation should be moved here
```
hardbyte: Remove this before note before merging

neilbacon: The new code has a test that serialized length (old code) is correct which fails for python 2, still deserializes correctly, but the serialized form isn't compatible between python 2 & 3. Haven't figured this out yet.

PR #193, opened 2018-09-14
A missing requirement and a backported import for TimeoutError

hardbyte:


clkhash/backports.py
```diff
@@ -108,6 +108,7 @@ def _p2_unicode_reader(unicode_csv_data, dialect=csv.excel, **kwargs):
         TimeoutError = globals()['__builtins__']['TimeoutError']
     except:
         pass    # pypy3 raises TypeError: 'module' object is not subscriptable
+        TimeoutError = TimeoutError
```
nbgl: Why is this necessary?

hardbyte: Because we need `TimeoutError` to be attached to the `clkhash.backports` module scope. :-/

Yes supporting all these Python versions is annoying.

nbgl: That makes sense. As far as I can tell though `TimeoutError = globals()['__builtins__']['TimeoutError']` will always throw, so the entire `try`/`except` block is redundant with the exception of `TimeoutError = TimeoutError`.

I think a neater way of rewriting lines 106-113 would be ```python
try:
    TimeoutError = TimeoutError
except NameError:
    # We're in Python < 3.3.
    TimeoutError = OSError
```

Comment (hardbyte) [reply clkhash/backports.py]: Oh wow, so we merged something that always throws? Opps!

I just assumed that on Python3.6+ that the module must have been subscriptable and pypy3 was just being a bit pedantic or annoying.

Comment (hardbyte) [reply clkhash/backports.py]: Oh and you can't just do `TimeoutError = TimeoutError` because that'd be too easy. (try it)

But this should work (I think) `globals()['__builtins__'].TimeoutError`

File: tests/test_backports.py [hunk 2]
```diff
@@ -7,10 +7,18 @@
 import io
 import unittest
 
+import clkhash.backports
 from clkhash.backports import (int_from_bytes, re_compile_full,
                                unicode_reader, strftime)
```
nbgl: Mixing import levels. Either use `import foo` or `from foo import bar`.

clkhash/backports.py
```diff
@@ -105,9 +105,10 @@ def _p2_unicode_reader(unicode_csv_data, dialect=csv.excel, **kwargs):
 
 if sys.version_info > (3, 2):
     try:
-        TimeoutError = globals()['__builtins__']['TimeoutError']
+        TimeoutError = globals()['__builtins__'].TimeoutError
     except:
-        pass    # pypy3 raises TypeError: 'module' object is not subscriptable
+        # pypy3 raises TypeError: 'module' object is not subscriptable
+        TimeoutError = TimeoutError
```
nbgl: You shouldn't need the `try`/`except` at all. If `TimeoutError` is in builtins then it will never throw. If it is not, then your except clause will throw.

clkhash/backports.py
```diff
@@ -105,9 +105,10 @@ def _p2_unicode_reader(unicode_csv_data, dialect=csv.excel, **kwargs):
 
 if sys.version_info > (3, 2):
     try:
-        TimeoutError = globals()['__builtins__']['TimeoutError']
+        TimeoutError = globals()['__builtins__'].TimeoutError
```
nbgl: No need to access `globals`. Use `__builtins__.TimeoutError`. Even better, just `TimeoutError` since it's a builtin.

clkhash/backports.py
```diff
@@ -105,9 +105,10 @@ def _p2_unicode_reader(unicode_csv_data, dialect=csv.excel, **kwargs):
 
 if sys.version_info > (3, 2):
```
nbgl: I suggest `try: TimeoutError = TimeoutError` instead of explicitly checking for versions. It's slightly safer as you don't have to keep track of the changelogs.

tests/test_backports.py
hardbyte: I think this is better as I want to test existence of particular attributes of the `backports` module which shouldn't be done at test import but rather at test execution - hence the `import clkhash.backports`. However the other backport tests are arguably cleaner by importing into the top level scope.

clkhash/backports.py
hardbyte: Good point. I've revised this to be much simpler - let's hope travis/jenkins agree.

clkhash/backports.py
hardbyte: Can't use the last one as it isn't a builtin in all versions but it is a good point that it was unnecessarily complex.

hardbyte:
Thanks for the review Jakub - I'll rebase and request you to have another look.

nbgl:
Looks like Mypy is failing. I think clkhash/backports.py:107 needs to change from
```python
    TimeoutError = __builtins__.TimeoutError
```
to
```python
    TimeoutError = __builtins__.TimeoutError  # type: type
```

Issue #205, opened 2018-10-12
Maybe redundant code using partial & __call__?


neilbacon:
In bloomfilter.py
class NgramEncodings(Enum):
   """ The available schemes for encoding n-grams.    ..
     the slightly awkward looking construction with the calls to partial
     and the overwrite of __call__ are due to
     compatibility issues with Python 2.7.
   """
Unfortunately this comment does not describe what the compatibility issues were, so its hard to know whether this class is still needed.

If I get rid of it (simply returning the encoding functions directly) the tests still run fine locally on ubuntu docker container with Python 2.7.15rc1.

PR #207, opened 2018-10-12
Redundant partial

neilbacon:
Remove class NgramEncodings(Enum) as per #205

### DIFF
--- a/clkhash/bloomfilter.py
+++ b/clkhash/bloomfilter.py
@@ -229,21 +229,20 @@ class NgramEncodings(Enum):
     def __call__(self, *args):
         return self.value(*args)
 
-    @classmethod
-    def from_properties(cls,
-                        properties  # type: GlobalHashingProperties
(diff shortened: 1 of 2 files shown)


nbgl:
@wilko77, you put that class in there in the first place. I recall you had some technical reasons for these, but I don’t remember what they were. What were they and are they still valid?

nbgl:
Thank you for this PR, Neil. I personally find this change to make for more elegant code, but I’d like to hear from Wilko regarding the above.

wilko77:
Back then, the `crypto_bloom_filter` had different arguments. Instead of the `hash_properties` you had do pass in the hash function directly. 
I chose to solve this with an enum of hash functions, as that way it is clear what the valid arguments are. And as a bonus you might get your editor to show the docstrings, which makes choosing the right one a breeze.

But since the API changed, this is not needed any more.

Issue #210, opened 2018-10-23
implementation of dynamic k can lead to bad approximation of n-gram similarity


wilko77:
The problem lies in the `get_k` function in the `FieldHashingProperties` class:

```
def get_k(self, num_ngrams):
        return (int(round(self.num_bits / num_ngrams)) if self.num_bits
else (self.k if self.k else 0))
```

The rounding here introduces inconsistencies as shown in the following example:

Let's say we have two entries for a feature `e_1` and `e_2`. 
Entry `e_1` consists of 66 tokens, `e_2` of 67, and they have 65 tokens in common.
Thus, the token-similarity is ~97.7%
We set `num_bits` for this feature to 100.
Then, the `k` value for `e_1` is 2, and 1 for `e_2`.
Thus, we create 132 indices for `e_1` and 67 indices for `e_2` to be inserted into the CLK.

In the best case (without any collisions), the CLK similarity can only be 130/199 ~ **65%**.

Another side-effect is that it increases the variance of the pop-count of the generated CLKs.

This problem is less pronounced the more `num_ngrams << num_bits`. However, in realistic settings, both values will be fairly close to each other.

I propose to use `variable k`, that is, we will always generate `num_bits` indices. If `num_bits` is not a multiple of the number of tokens, then we create different `k` values for the tokens, by "spreading" the remainder over the first few tokens such that the maximum difference of the k values is 1.

In our example from above, the first 34 tokens in `e_1` will be encoded with k=2 and the rest with k=1.
Now, if we look at the best case similarity of the generated CLKs again, we have to consider if the non-matching token was encoded with k=1 or k=2.
For k=1 the similarity is 98% vs. 96% for k=2.

neilbacon:
Ah, we did have a little discussion of the pros & cons of keeping num_bits exact as in your code, but it looks like it was too little. Shall I change get_k() to get_ks() returning a list of k values as in your code?

neilbacon:
Fixed by pr #211

PR #211, opened 2018-10-24
use array of k's to fix bug #210

neilbacon:
Addresses #210


neilbacon:
Thanks. I'll try to add a test corresponding to what is in #210 as well.

neilbacon:
Thanks. Done and added a test corresponding to what is in #210 as well.

PR #235, opened 2019-02-28
Feature auto deploy

hardbyte:
The idea is that one of the CI systems can create and publish the artifacts to make releases more reproducible. After this PR I'll look at doing the same for GitHub releases.

Docs links:
- https://packaging.python.org/guides/supporting-windows-using-appveyor/
- https://www.appveyor.com/docs/deployment/


hardbyte:
I think the auto version bump is a bit harder - at least with the version stored in a file as we have it.
For now we just have to ensure when me merge something into master it has already had a version bump.

PR #241, opened 2019-04-15
Test update urls

hardbyte:


hardbyte:
Ok, let's use those changes

PR #247, opened 2019-05-09
Fix #246

### LINKED ISSUE #246: clkutil crashes during hashing when schema defines 'blakeHash'
The problem seems to be that `clkutil` enforces the user to provide two secrets, whereas the Blake hash scheme only needs one. 
Internally, we expand the two secrets to a list of two-tuples, one for each feature. 
The Blake hash encoding code in `bloomfilter.py` gets fed the keys as a two-tuple, and then tries to unpack with `key, = keys`.

wilko77:
fixes issue #246

The underlying problem is that 
- Blake hash requires one key, whereas double hash needs two
- we can define the hash type on a per field basis
- clkutil enforces users to input two secrets

The Blake hash function breaks if called with two secrets. However, you cannot avoid calling it with two keys, if you have 
- mixed hash types in the schema
- used clkutil

This is a quick fix. We should probably rethink the key derivation such that we generate only the keys needed from _one_ secret.

### DIFF
--- a/clkhash/bloomfilter.py
+++ b/clkhash/bloomfilter.py
@@ -209,7 +209,7 @@ def blake_encode_ngrams(ngrams,   # type: Iterable[str]
         :return: bitarray of length l with the bits set which correspond to the
                  encoding of the ngrams
     """
-    key, = keys  # Unpack.
+    key = keys[0]  # We only need the first key
 
     log_l = int(math.log(l, 2))
     if not 2 ** log_l == l:
--- a/tests/test_e2e_hashing.py
+++ b/tests/test_e2e_hashing.py
@@ -179,3 +179,62 @@ class TestHashingWithDifferentK(unittest.TestCase):
         self.assertGreater(n2, n15)
         self.assertLessEqual(n15, round(n1 * 1.5))
         self.assertLessEqual(n2, n1 * 2)
+
+
+class TestHashingWithDifferentHashFunctions(unittest.TestCase):
+    """Schema allows to define the hash function on a per field basis."""
+
+    def test_different_hash_functions(self):
+        schema = Schema(
+            l=1024,
+            xor_folds=0,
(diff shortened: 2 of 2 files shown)


clkhash/bloomfilter.py
```diff
@@ -209,7 +209,7 @@ def blake_encode_ngrams(ngrams,   # type: Iterable[str]
         :return: bitarray of length l with the bits set which correspond to the
                  encoding of the ngrams
     """
-    key, = keys  # Unpack.
+    key = keys[0]  # We only need the first key
```
hardbyte: The docstring says `:param keys: secret key for blake2 as bytes` and the type signature says `type: Sequence[bytes]`

The first index of the *documented* bytes argument would be an int in Python3 and a str in Python2... Any chance of using a one character key by accident seems... suboptimal.

Fix is just updating the docstring as with mypy we are fairly certain we are calling with a sequence of bytes.

wilko77: We use this definition throughout the codebase for two-tuples of keys. I opened issue #248 to discuss if the current type definition is the best choice.
I don't think this should be part of this PR.
>The first index of the documented bytes argument would be an int in Python3...

That's not quite right. See #248.

hardbyte: > We use this definition throughout the codebase for two-tuples of keys. I opened issue #248 to discuss if the current type definition is the best choice.
> I don't think this should be part of this PR.

 I think the type definition is fine and agree it shouldn't be changed here. What I found misleading is the docstring doesn't line up with the type def. To my reading the docstring suggests a bytestring could be the argument (let's ignore Python 2 being silly in this regard), and only in that case e.g. when a `str` or `bytes` object is passed in would the next statement be true: 

> > The first index of the documented bytes argument would be an int in Python3...
> 
> That's not quite right. See #248.

PR #250, opened 2019-05-26
Support multi-party linkage in the Entity Service

nbgl:
This is a change to the CLI. The rest client module already supported multi-party linkage.


nbgl:
The tests are failing because they are being run against an older deployment of the Entity Service, which does not support multiparty linkage.

PR #251, opened 2019-05-28
Travis optional server fail

hardbyte:


hardbyte:
Thanks Wilko. I agree that I'd prefer to **require** the integration tests to pass. In this case (the target branch adds multiparty support) the functionality isn't yet part of a release of the anonlink-entity-service - see https://github.com/data61/anonlink-entity-service/issues/372 we could simply say this is blocked until that is complete.

> All the integration tests will run the unit tests again

Yeah I realized that but at least every job within a stage runs in parallel so it shouldn't add too much to the CI time. If it is annoying we should look at separating the unit tests and integration tests better.

> we should change the 'SKIP_TEST_NB' switch from disabler to enabler, in order to de-clutter.

Done

hardbyte:
Unblocked with new release and deployment of anonlink-entity-service, have removed allowed failure of integration test stage.

.travis.yml
```diff
@@ -1,60 +1,79 @@
...
     - os: osx
       osx_image: xcode8.3
       python: "3.6-dev"
+
+    - stage: Integration
+      python: '3.8-dev'
       env:
+        - TEST_ENTITY_SERVICE=https://testing.es.data61.xyz
+        - INCLUDE_CLI=1
         - SKIP_TEST_NB=1
```
wilko77: delete

PR #255, opened 2019-06-07
Fixes to schema v2

hardbyte:
Schema v2 could have been better


hardbyte:
Note the tests are failing because of the tutorials - that will be addressed in https://github.com/data61/clkhash/pull/257

clkhash/schema.py
```diff
@@ -254,20 +296,12 @@ def validate_schema_dict(schema):
...
-    master_schema_bytes = _get_master_schema(version)
-    try:
-        master_schema = json.loads(master_schema_bytes.decode('utf-8'))
-    except ValueError as e:  # In Python 3 we can be more specific with
-        # json.decoder.JSONDecodeError, but that
-        # doesn't exist in Python 2.
-        msg = ('The master schema is not a valid JSON file. The schema cannot '
-               'be validated. Please file a bug report.')
-        raise_from(MasterSchemaError(msg), e)
+    master_schema = _get_master_schema(version)
```
hardbyte: Moved this bytes -> json handling into the helper function.

clkhash/schemas/v2.json
```diff
@@ -136,6 +141,7 @@
     },
     "hashingConfig": {
       "type": "object",
+      "additionalProperties": false,
```
hardbyte: This would have caught the invalid schemas.

hardbyte:
I have a local branch which updates the docs further - as https://clkhash.readthedocs.io/en/latest/schema.html refers to v1

clkhash/field_formats.py
```diff
@@ -169,11 +171,18 @@ def fhp_from_json_dict(
             always set to the default value.
         :return: A :class:`FieldHashingProperties` instance.
     """
+
+    # Loading default values directly from v2 master schema
+    from clkhash.schema import _get_master_schema
+    master_schema = _get_master_schema(version=2)
+    default_hashing_strategy = master_schema['definitions']['hashingConfig']['properties']['strategy']['default']
```
wilko77: I don't think a default for strategy makes sense. This is the most important value for tuning the schema to achieve good linkage. We should enforce users to take it seriously. Also, there is a dependency between clk length, number of features and strategy. A fixed strategy, which doesn't take the number of features into account is pretty useless most of the time.

Furthermore, for all other default values, we ignore what's written in the schema spec, but have something either hardcoded in code (e.g. schema.py line 184), default argument values (e.g. field_formats.py line 92, 93), or some hidden configurable variables (e.g. field_formats.py line 85,86). Adding default values from schema extraction, ahem, celebrates diversity and is thus in line with our company values, however, it also might be the last straw that tips the next review over the edge of despair. Maybe we should start spreading links to the EAP around... just in case.

On a more serious note, we should rethink all this defaults business. I don't like, that:
- the defaults are spread throughout the code base
- there is no guarantee that the default in the schema correlates to the default in the code
- the user doesn't actually know the exact details of the schema used to create the CLKs. This makes it unnecessarily hard to reproduce results...
But that should probably go into a separate issue.

Nevertheless, instead of this hardcoded default extraction, we should consider doing it like this https://python-jsonschema.readthedocs.io/en/stable/faq/#why-doesn-t-my-schema-s-default-property-set-the-default-on-my-instance. This way, it would not just be more elegant, but also, it would create a json structure with all values set, which we could then feed back to the user -> the effective schema. Wow.

clkhash/field_formats.py
```diff
@@ -169,11 +171,18 @@ def fhp_from_json_dict(
...
+
+    hashing_strategy = json_dict.get('strategy', default_hashing_strategy)
     h = json_dict.get('hash', {'type': 'blakeHash'})
-    num_bits = json_dict.get('numBits')
-    k = json_dict.get('k')
-    if not num_bits and not k:
-        num_bits = 200 # default for v2 schema
+
+    num_bits = hashing_strategy.get('numBits')
+    k = hashing_strategy.get('k')
```
wilko77: oh my, I really don't like that you can instantiate `FieldHashingProperties` with both `k` and `num_bits`.
Strategy should be its own thing that's either fixed_k or (exclusive) fixed_num_bits. I guess another issue...

clkhash/field_formats.py
```diff
@@ -843,14 +853,20 @@ def spec_from_json_dict(
...
-    """ Turns a dictionary into the appropriate object.
+    """ Turns a dictionary into the appropriate FieldSpec object.
 
         :param dict json_dict: A dictionary with properties.
+        :raises InvalidEntryError:
         :returns: An initialised instance of the appropriate FieldSpec
             subclass.
     """
-    if 'ignored' in json_dict:
+    if 'ignored' in json_dict and json_dict['ignored']:
```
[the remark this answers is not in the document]
wilko77: that cannot happen if you validate against schema. We enforce `True` with the `const` thing.

clkhash/schemas/v2.json
wilko77: or we could modify line 145 to `"required": ["ngram", "strategy"],`
as it doesn't really make sense to have a default here anyway.

docs/schema.rst
```diff
@@ -16,8 +16,10 @@ The hashing-schema is a detailed description of exactly what is fed to the hashi
 along with any configuration for the hashing itself.
 
 The format of the hashing schema is defined in a separate ``JSON Schema`` document
-`schemas/v1.json <https://github.com/data61/clkhash/blob/master/clkhash/schemas/v1.json>`_.
+`schemas/v2.json <https://github.com/data61/clkhash/blob/master/clkhash/schemas/v2.json>`_.
 
+Earlier versions of the hashing schema will continue to work, internally they are converted to
+``v2``.
```
wilko77: what about all the stuff underneath that still describes schema v1?

clkhash/cli.py
```diff
@@ -318,6 +322,28 @@ def generate_default_schema(output):
...
+    try:
+        clkhash.schema.from_json_file(
+            schema_file=schema,
+            validate=True
+        )
+
+        log("schema is valid", color='green')
+
+    except SchemaError as e:
+        log(str(e))
```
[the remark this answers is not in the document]
hardbyte: `jsonschema` just outputs the first error - although with a small tweak it could iterate over them. https://python-jsonschema.readthedocs.io/en/stable/validate/#jsonschema.IValidator.iter_errors

clkhash/field_formats.py
hardbyte: I've removed the `strategy` default and created https://github.com/data61/clkhash/issues/258 to discuss a unified way of handling default values.

clkhash/schemas/v2.json
```diff
@@ -144,9 +150,26 @@
           ],
           "description": "this construction is for future proofing. We might want to parameterize the hash functions one day..."
         },
-        "strategy": {"type": "object", "default": {"numBits": 200}, "oneOf": [
-          {"numBits": {"type": "number", "minimum": 1, "default": 200, "description": "dynamic k = numBits / number of n-grams"}},
-          {"k": {"type": "number", "minimum": 1, "default": 20, "description": "max number of bits per n-gram"}}
+        "strategy": {"type": "object", "default": {"numBits": 200}, "oneOf":
```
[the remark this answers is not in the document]
hardbyte: Agree, https://github.com/data61/clkhash/issues/259

clkhash/schemas/v2.json
```diff
@@ -144,9 +150,26 @@
...
+              "description": "dynamic k = numBits / number of n-grams"
+            }
+          },
+          {
+            "required": ["k"],
+            "k": {
+              "type": "number",
+              "minimum": 1,
+              "default": 20,
+              "description": "max number of bits per n-gram"
```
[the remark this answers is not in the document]
hardbyte: Also part of #259

clkhash/field_formats.py
hardbyte: Yup the `FieldHashingProperties` is currently suboptimal... https://github.com/data61/clkhash/issues/260

hardbyte:
Thanks for the review Wilko. I've created a bunch of issues and tried not to adjust the schema itself in this PR. I wanted to mostly tighten the spec up and ensure all our examples/tests were compliant.

Shame I didn't pick up on the strategy issue back [here](https://github.com/data61/clkhash/pull/201/files#r224629218)

Only one thing you said I had an issue with:
> that cannot happen if you validate against schema. We enforce `True` with the const thing.
I think that `ignored=False` should be valid for any feature so someone can easily toggle a feature on and off.

wilko77:
I see. Good point. 
So an `ignoreFeature` section needs to have `ignored: true`, but any other `featureConfig` can have an optional `ignored: true/false`.
We should probably make this explicit in the `featureConfig` definition and also test for the correct behavior.

clkhash/bloomfilter.py
```diff
@@ -319,9 +319,10 @@ def crypto_bloom_filter(record,      # type: Sequence[Text]
         if fhp:
             ngrams = list(tokenize(field.format_value(entry)))
             hash_function = hashing_function_from_properties(fhp)
-            bloomfilter |= hash_function(ngrams, key,
-                                         fhp.ks(len(ngrams)),
-                                         hash_l, fhp.encoding)
+            if ngrams:
```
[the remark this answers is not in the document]
wilko77: `fhp.ks` should handle a division by zero by itself. It doesn't state that zero is not a valid input.
However, I like you `ngrams` test, as it will avoid unnecessary calls to 'hash_function' if no ngrams are present.
-> Good execution, wrong justification ;-)

wilko77:
as you very elegantly diverted all my concerns into separate issues, you could sent the PR again for me to sign off.

docs/schema.rst
```diff
@@ -219,19 +208,60 @@ name          type                     optional description
 ============  ======================   ======== ===========
 ngram         integer                  no       specifies the n in n-gram (the tokenization of the input values).
 positional    boolean                  yes      adds the position to the n-grams. String "222" would be tokenized (as uni-grams) to "1 2", "2 2", "3 2"
-weight        float                    yes      positive number, which adjusts the number of hash functions (k) used for encoding. Thus giving this feature more or less importance compared to others.
+strategy      :ref:`schema/strategy`   yes      the strategy for assigning bits to the encoding
```
wilko77: strategy is not optional any more in the schema definition.

requirements.txt
```diff
@@ -5,7 +5,7 @@ cryptography==2.6.1
 enum34==1.1.6; python_version < '3.4'
 future==0.17.1
 futures==3.2.0; python_version < '3.2'
-jsonschema==2.6.0
+jsonschema==3.0.1
```
hardbyte: Something, quite possibly this, has broken the pyinstaller - https://ci.appveyor.com/project/hardbyte/clkhash/history

hardbyte: Reverting back to jsonschema 2.6 seems to have fixed the windows exe build - https://ci.appveyor.com/project/hardbyte/clkhash/builds/25527663

PR #257, opened 2019-06-09
Update tutorial notebook

hardbyte:
Note this is based on `fix-schema-v2` branch, the aim being to update both tutorial notebooks to use the "new" v2 schema. The notebooks told the reader to try with different thresholds/schemas and I've expanded slightly to do that so the tutorial is more complete when not being interactively executed.

There were also calls to a deprecated `anonlink` api which I've updated, and for good measure I've split the notebooks tests into their own task within the integration test stage in travis.

Finally I've started to work towards the notebooks being fully tested as opposed to executing without any cells throwing errors. ~This isn't yet complete and I haven't removed the `lax` command yet.~


hardbyte:
Looks good - I'll rebase and merge back into the schema v2 branch if/once tests are passing

PR #261, opened 2019-06-19
Testing ignored

wilko77:
I wrote some tests to check if the handling of 'Ignored' features is correct.
More specifically, I want good failures.
Currently one test fails, because field_formats raises an `InvalidSchemaError`, caused by `KeyError`.
I argue that this should be an `InvalidEntryError`. (see descriptions of InvalidSchemaError and InvalidEntryError). 

More generally, I want to be allowed:
```
{
    "identifier": "good",
    "ignored": true
}
```
```
{
   "identifier": "good_with_format",
    "ignored": true,
    "format": {
        "type": "date", "format": "%Y-%m-%d"},
        "hashing": {"ngram": 1, "strategy": {"k": 20}}
}
```
```
{
   "identifier": "good_with_format",
    "ignored": false,
    "format": {
        "type": "date", "format": "%Y-%m-%d"},
        "hashing": {"ngram": 1, "strategy": {"k": 20}}
}
```
but not
```
{
    "identifier": "good",
    "ignored": false
}
```

Once we agree on the correct handling, I'll write the corresponding fix. 
So for now, just comment (dis)agreement.

### DIFF
--- a/tests/test_field_formats.py
+++ b/tests/test_field_formats.py
@@ -465,3 +465,36 @@ class TestFieldFormats(unittest.TestCase):
         self.assertEqual(spec.hashing_properties.missing_value.replace_with,
                          spec.hashing_properties.replace_missing_value(''))
 
+    def test_ignored(self):
+        spec_dict = {
+            'identifier': 'testingIgnored',
+            'ignored': True}
+        spec = field_formats.spec_from_json_dict(spec_dict)
+        # clkhash handles ingored fields by setting hashing properties to None.
+        self.assertIsNone(spec.hashing_properties)
+        self.assertEqual(spec.identifier, 'testingIgnored')
+        spec_dict = {
+            'identifier': 'testingIgnored',
+            'ignored': False}
+        with self.assertRaises(field_formats.InvalidEntryError):
+            field_formats.spec_from_json_dict(spec_dict)
+        spec_dict = {
+            'identifier': 'ignoredDates',
+            'ignored': True,
+            'format': {
+                'type': 'date', 'format': '%Y-%m-%d'},
+            'hashing': {'ngram': 0, 'strategy': {'k': 20}}
+        }
+        spec = field_formats.spec_from_json_dict(spec_dict)
+        self.assertIsNone(spec.hashing_properties)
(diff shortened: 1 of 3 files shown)


tests/test_field_formats.py
```diff
@@ -465,3 +465,36 @@ def test_enum(self):
...
+        spec_dict = {
+            'identifier': 'notIgnoredDates',
+            'ignored': False,
+            'format': {
+                'type': 'date', 'format': '%Y-%m-%d'},
+            'hashing': {'ngram': 0, 'strategy': {'k': 20}}
+        }
+        spec = field_formats.spec_from_json_dict(spec_dict)
+        self.assertIsNotNone(spec.hashing_properties)
+        self.assertEqual(spec.identifier, 'notIgnoredDates')
```
hardbyte: Suggest adding a check that `spec` is an instance of `DateSpec`

tests/test_schema.py
```diff
@@ -96,6 +97,24 @@ def test_good_schema2_repr(self):
...
+                'kdf': {
+                    'type': 'HKDF'}},
+            'features': [
+                {
+                    'identifier': 'rec_id',
+                    'ignored': False}]
+        }
+        try:
+            schema.from_json_dict(schema_dict)
+        except Exception as e:
```
hardbyte: Better to use:
```python
with self.assertRaises(Exception) as contextmanager:
    schema.from_json_dict(schema_dict)

exception = contextmanager.exception
self.assertIsInstance(exception.__cause__,  ValidationError)
```

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

tests/test_schema.py
```diff
@@ -96,6 +97,24 @@ def test_good_schema2_repr(self):
         assert "v2" in schema_repr
         assert "12 fields" in schema_repr
 
+    def test_validation_of_invalid_ignored(self):
```
hardbyte: It isn't really an invalid ignored feature is it? It is invalid because it is NOT ignored and is illdefined because it doesn't comply with the schema of any type of feature specification.

PR #265, opened 2019-07-01
Deletion from command line

hardbyte:
Added project and run deletion commands to the command line. 

App Veyor deployment now handles existing uploads (to prepare for switching to Azure DevOps)

Updates to latest requirements. 
Note the latest version of `requests` didn't work for Python 3.4, as we don't officially support 3.4 anymore I removed them from jenkins and appveyor.

I tweaked the rest_client tests to not hash the CLK data for every single test.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,7 +2,7 @@ environment:
   global:
     TWINE_USERNAME: confidentialcomputing
     TWINE_PASSWORD:
-      secure: 8iJVrN3C/NidiVaa4MArSaNepgPz/AnBn2AC6i0EomM=
+      secure: v/CA7epn+w6NgHihL5rLRc8+lUbgNMsMsMp5/XPWSoI=
 
   INCLUDE_CLI: 1
 
@@ -61,6 +61,6 @@ artifacts:
   - path: dist\*
 
 deploy_script:
-  - ps: if ($env:APPVEYOR_REPO_TAG -eq $TRUE) { pip install twine; twine upload dist/* }
+  - ps: if ($env:APPVEYOR_REPO_TAG -eq $TRUE) { pip install twine; twine upload --skip-existing dist/* }
 
 deploy: on
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -14,9 +14,11 @@ environment:
     - PYTHON: "C:\\Python27"
     - PYTHON: "C:\\Python35"
     - PYTHON: "C:\\Python36"
+    - PYTHON: "C:\\Python37"
     - PYTHON: "C:\\Python27-x64"
     - PYTHON: "C:\\Python35-x64"
     - PYTHON: "C:\\Python36-x64"
+    - PYTHON: "C:\\Python37-x64"
 
 install:
   # Check that we have the expected version and architecture for Python
@@ -59,6 +61,6 @@ artifacts:
   - path: dist\*
 
 deploy_script:
-  - ps: if ($env:APPVEYOR_REPO_TAG -eq $TRUE) { pip install twine; twine upload --skip-existing dist/* }
+  - ps: if ($env:APPVEYOR_REPO_TAG -eq $TRUE) { pip install twine; twine upload --skip-existing dist/*.whl }
 
-deploy: on
+deploy: off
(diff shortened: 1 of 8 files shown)


clkhash/cli.py
```diff
@@ -279,6 +279,50 @@ def results(project, apikey, run, watch, server, output):
         log("No result yet")
 
 
+@cli.command('delete', short_help="delete a run on the entity service")
+@click.option('--server', type=str, default=DEFAULT_SERVICE_URL, help="Server address including protocol")
+@click.option('--project', help='Project identifier')
+@click.option('--run', help='Run ID to get results for')
```
gusmith: Run to delete?

clkhash/cli.py
```diff
@@ -279,6 +279,50 @@ def results(project, apikey, run, watch, server, output):
...
+        if verbose:
+            log(msg)
+    except ServiceError as e:
+        log("Unexpected response with status {}".format(e.status_code))
+        log(e.text)
+    else:
+        log("Run deleted")
+
+
+@cli.command('delete-project', short_help="create a run on the entity service")
```
gusmith: Wrong copy/paste of the short help :)

clkhash/cli.py
```diff
@@ -279,6 +279,50 @@ def results(project, apikey, run, watch, server, output):
...
+@click.option('--server', type=str, default=DEFAULT_SERVICE_URL, help="Server address including protocol")
+@click.option('--project', help='Project identifier')
+@click.option('--apikey', type=str, help="Project Authorization Token")
+@click.option('-v', '--verbose', default=False, is_flag=True, help="Script is more talkative")
```
gusmith: Just a question, if you delete a project, I would assume you are also deleting all the runs in the project.
If this is the case, do you want some safeguards: 
 - you can delete a project if it has no runs
 - otherwise you have to force delete?

tests/test_cli.py
```diff
@@ -492,6 +509,40 @@ def test_create_project_and_run(self):
...
+            'delete-project',
+            '--server', self.url,
+            '--project', project['project_id'],
+            '--apikey', project['result_token']
+        ]
+        cli_result = runner.invoke(clkhash.cli.cli, command)
+        assert cli_result.exit_code == 0, cli_result.output
+
+        with pytest.raises(ServiceError):
+            rest_client.project_get_description(self.url, project['project_id'], project['result_token'])
```
gusmith: >       with pytest.raises(ServiceError):
>           rest_client.project_get_description(self.url, project['project_id'], project['result_token'])

Can you do the same for the `test_delet_run`?

.appveyor.yml
```diff
@@ -12,11 +12,9 @@ environment:
     # http://www.appveyor.com/docs/installed-software#python
 
     - PYTHON: "C:\\Python27"
-    - PYTHON: "C:\\Python34"
     - PYTHON: "C:\\Python35"
     - PYTHON: "C:\\Python36"
```
gusmith: No Python 3.7?

tests/test_rest_client.py
```diff
@@ -20,33 +20,60 @@
...
+    @classmethod
+    def _delete_created_projects(cls):
+        for project in cls._created_projects:
+            try:
+                rest_client.project_delete(cls.url,
+                                           project['project_id'],
+                                           project['result_token']
+                                           )
+            except ServiceError:
+                pass
```
gusmith: Do you want to log something if an error is raised, or it may happen and it is not important?
Would it be the same comment as in another class where you wrote: "probably already deleted".

clkhash/cli.py
hardbyte: Valid comment but I'm going to keep the current behavior.

.appveyor.yml
hardbyte: Good point, I'll add it and see if it works. I'm keen not to add more work to AppVeyor though

clkhash/cli.py
gusmith: :+1:

.appveyor.yml
gusmith: Good point.

PR #267, opened 2019-07-04
Remove the triggers excluding branches.

### LINKED ISSUE #266: Azure pipeline not build master branch
This is a misconfiguration of the the triggers in the azure pipeline file.
From Azure documentation:

> Includes are processed first, and then excludes are removed from that list.

So
```
trigger:
  branches: 
    include:
    - master
    eclude:
    - '*'
```
was setting the list to include to master, and then removing it.

gusmith:
Fix #266


gusmith:
The build on travis is failing because of a flaky test. I opened the issue #272 tracking it. 
But for this PR, I cannot merge it because of this test, and I cannot restart a build on travis.

PR #270, opened 2019-07-04
Fix docs

gusmith:
The documentation is now building on [readthedocs](https://clkhash.readthedocs.io/en/fix-docs/index.html)


docs/tutorial_api.ipynb
```diff
@@ -19,7 +19,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
```
hardbyte: I think the notebook tests are most useful if we commit outputs? (Because they check for consistency)

gusmith: I didn't know the notebook tests were checking for consistency.
I only did it because I remembered the commit 0fdf3d1639e37a812fa672901201719ee2306d7b merged in the PR #265 in which all the outputs from the `tutorial_cli.ipynb` had been removed.

gusmith: Should I bring back the outputs for both notebooks or keep them empty?
Note: the notebook are run by readthedocs, including the output in the documentation.

hardbyte: Yeah sorry that was a mistake in #265 - if you could run and commit the output that would be good.

gusmith: :+1:

Issue #98, opened 2018-04-13
Attribute the first/last name dataset source


hardbyte:
Aha! Link: https://csiro.aha.io/features/ANONLINK-40

hardbyte:
Closed in #271

PR #271, opened 2019-07-04
Add Distributions to PII Generation

ProgrammedByJake:
Added a way to allow distributions to be used for PII generation. First names, last names and birthdays are all now distributed based on the counts provided in the census data.

From my own testing it slows down generation of PII to around 0.32x what it once was, but it was so marginal in the scheme of the benchmark that I'm not sure how worried we are regarding the performance of this area. 

With the new data I have provided, all documents now have a reference in code, I don't know if they should be displayed anywhere else. #98 

EDIT: Automated testing is clearly something I need to learn more about. I will attempt to generate some thorough unit tests.


clkhash/randomnames.py
```diff
@@ -61,20 +45,87 @@ def save_csv(data,  # type: Iterable[Tuple[Union[str, int], ...]]
...
+            for row in reader:
+                try:
+                    self.total += int(row[1])
+                except ValueError:
+                    raise ValueError("Distribution resources must only contain integers in the second column.")
+                self.indices.append(self.total)
+                self.values.append(row[0])
+
+    def generate(self):
+        # type: () -> str
```
hardbyte: Why return a `str` here?

clkhash/randomnames.py
```diff
@@ -61,20 +45,87 @@ def save_csv(data,  # type: Iterable[Tuple[Union[str, int], ...]]
...
+        """ Generates a random value, weighted by the known distribution
+        """
+
+        target = random.randint(0, self.total)
+        return self.binary_search(0, len(self.values) - 1, target)
+
+    def binary_search(self, l, h, t):
+        # type: (int, int, int) -> str
+        """ Return the value of the index between 0 and
```
hardbyte: We can use builtins to achieve this - see the bisect module https://docs.python.org/3.7/library/bisect.html#module-bisect

Also not sure why returning a `str`? 
docstring is incomplete

ProgrammedByJake: Thanks for the guidance, I've implemented the bisect method rather than reinventing the wheel.

hardbyte:
I'll merge this now, but it would be most appreciated if you did want to open a new PR to test the distributions code.

PR #274, opened 2019-07-05
Fix tokenizer

wilko77:
There is a bug in the tokenizer.

If a feature is encoded as bi-grams and a value happens to be empty, then the tokenizer would return a token of two white-spaces. However, the correct behavior would be to return nothing.

This is especially bad for the *variable-k* encoding scheme:
A missing value, as it will be tokenized into a "  " token, would be encoded with *numBits* bits. Thus all entities with missing values for that feature will have all those bits in common and in turn an inflated similarity score. This screws up the matching pretty badly!

### DIFF
--- a/clkhash/tokenizer.py
+++ b/clkhash/tokenizer.py
@@ -56,6 +56,9 @@ def get_tokenizer(fhp  # type: Optional[field_formats.FieldHashingProperties]
         if ignore is not None:
             word = word.replace(ignore, '')
 
+        if len(word) == 0:
+            return tuple()
+
         if n > 1:
             word = ' {} '.format(word)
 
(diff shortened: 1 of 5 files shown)


hardbyte:
Due to a change in the CLKs in the tutorial we will have to rerun and save the output of `tutorial_api.ipynb`

https://travis-ci.org/data61/clkhash/jobs/554551004

hardbyte:
Ahh right sorry I didn't anticipate that both notebooks would need rerunning.

This latest failure is the version - might as well remove the beta part and use `0.13.1` - I'll edit now

gusmith:
@hardbyte The last commit (resolving conflicts with the master branch merge) solve the issues related to the notebook tests. I guess it can be merged in now :)

PR #275, opened 2019-07-05
Add test cases for distributions

ProgrammedByJake:
Added test cases for the new distributions for PII.

@hardbyte this is my first time with pytest based testing, so if you have any pointers I would be most thankful. I managed to create tests for all bar 2 lines.


tests/test_names.py
```diff
@@ -10,6 +15,45 @@
...
+
+        for i in range(1000):
+            self.assertGreaterEqual((rn.random_date(2018, age_dist)-start).days, 0)
+            self.assertLess((rn.random_date(2018, age_dist)-end).days, 0)
+
+    def test_distribution(self):
+        # No distribution
+        age_dist = None
+        with pytest.raises(ValueError):
+            rn.random_date(2018, age_dist)
```
hardbyte: This is testing how `random_date` handles a missing distribution. Should be in its own test (or part of the random date tests)

tests/test_names.py
```diff
@@ -10,6 +15,45 @@
...
+        # Ensure dates fall in range
+        age_dist = rn.Distribution('../tests/testdata/ages.csv')
+        start = datetime(2016, 1, 1)
+        end = datetime(2019, 1, 1)
+
+        for i in range(1000):
+            self.assertGreaterEqual((rn.random_date(2018, age_dist)-start).days, 0)
+            self.assertLess((rn.random_date(2018, age_dist)-end).days, 0)
+
+    def test_distribution(self):
```
hardbyte: `test_distribution` should include calling `Distribution.generate()` and checking the output is as expected.

ProgrammedByJake:
Thanks again for the help @hardbyte

PR #277, opened 2019-07-08
The client retry its requests if receiving a `RateLimitedClient` error.

### LINKED ISSUE #272: test_cli has an unreliable test
The test `tests.test_cli.TestCliInteractionWithService testMethod=test_2_party_upload_and_results` fail regularly. Example:
https://magic-ci-atp.it.csiro.au/job/clkhash/job/fix-issue-266/1/console
But passing when rebuilding:
https://magic-ci-atp.it.csiro.au/job/clkhash/job/fix-issue-266/2/console

It also start failing on Travis:
https://travis-ci.org/data61/clkhash/builds/554029404?utm_source=github_status&utm_medium=notification

```
=================================== FAILURES ===================================
________ TestCliInteractionWithService.test_2_party_upload_and_results _________
self = <tests.test_cli.TestCliInteractionWithService testMethod=test_2_party_upload_and_results>
    def test_2_party_upload_and_results(self):
        project, run = self._create_project_and_run()
    
        def get_coord_results():
            # Get results from coordinator
            return self.run_command_capture_output(
                [
                    'results',
                    '--server', self.url,
                    '--project', project['project_id'],
                    '--run', run['run_id'],
                    '--apikey', project['result_token']
                ]
            )
    
        # Upload Alice
        alice_upload = self.run_command_load_json_output(
            [
                'upload',
                '--server', self.url,
                '--project', project['project_id'],
                '--apikey', project['update_tokens'][0],
                self.clk_file.name
            ]
        )
        self.assertIn('receipt_token', alice_upload)
    
        out_early = get_coord_results()
        self.assertEqual("", out_early)
    
        # Upload Bob (subset of clks uploaded)
        bob_upload = self.run_command_load_json_output(
            [
                'upload',
                '--server', self.url,
                '--project', project['project_id'],
                '--apikey', project['update_tokens'][1],
                self.clk_file_2.name
            ]
        )
    
        self.assertIn('receipt_token', bob_upload)
    
        # Give the server a small amount of time to process
        time.sleep(5.0)
    
        results_raw = get_coord_results()
>       res = json.loads(results_raw)
tests/test_cli.py:636: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/python/3.7.1/lib/python3.7/json/__init__.py:348: in loads
    return _default_decoder.decode(s)
/opt/python/3.7.1/lib/python3.7/json/decoder.py:337: in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <json.decoder.JSONDecoder object at 0x7ff2b4811e80>, s = '', idx = 0
    def raw_decode(self, s, idx=0):
        """Decode a JSON document from ``s`` (a ``str`` beginning with
        a JSON document) and return a 2-tuple of the Python
        representation and the index in ``s`` where the document ended.
    
        This can be used to decode a JSON document from a string that may
        have extraneous data at the end.
    
        """
        try:
            obj, end = self.scan_once(s, idx)
        except StopIteration as err:
>           raise JSONDecodeError("Expecting value", s, err.value) from None
E           json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
/opt/python/3.7.1/lib/python3.7/json/decoder.py:355: JSONDecodeError
```

gusmith:
Fix #272 

The tests for the cli are relying on a real anonlink-entity-service to deployed. But the nginx proxy of the service is limiting the number of requests it will accept per second (limitation of request per second and burst of request).

The `clkhash` cli was not limiting anything and was not retrying on failure.

To solve this, we use the [`retrying`](https://pypi.org/project/retrying/) package. If a `RateLimitedClient` error is raised, the client will wait 1 second before retrying, and will retry up to 10 times.


clkhash/rest_client.py
```diff
@@ -4,9 +4,13 @@
 import clkhash
 from clkhash.backports import TimeoutError
 import logging
+from retrying import retry
 
 logger = logging.getLogger(__name__)
 
+WAITING_TIME_WHEN_RATE_LIMITED_MS = 2000  # 1 seconds
```
hardbyte: This looks like 2 seconds
```suggestion
WAITING_TIME_WHEN_RATE_LIMITED_MS = 2000
```

clkhash/rest_client.py
```diff
@@ -24,6 +28,11 @@ class RateLimitedClient(ServiceError):
     """
 
 
+def retry_if_reate_limited_error(exception):
```
hardbyte:
```suggestion
def _is_rate_limited_error(exception):
```

Type in *reate* and (despite the retry docs) it is more pythonic to use `is_...`

clkhash/rest_client.py
```diff
@@ -35,6 +44,8 @@ def _handle_json_response(response, failure_message, expected_status_code=200):
         raise ServiceError(failure_message, response)
 
 
+@retry(wait_fixed=WAITING_TIME_WHEN_RATE_LIMITED_MS, stop_max_attempt_number=MAX_NUMBER_REQUEST_RETRIES,
```
hardbyte: I'd prefer to use the exponential back-off options instead of trying again after 1s 10 times in a row.

Perhaps:
`wait_exponential_multiplier=100, wait_exponential_max=10000` (with the numbers being configurable)

We are also repeating the same thing with the `@retry` decorator - consider pulling out all the requests calls to a single location where the retry can be applied once.

setup.py
```diff
@@ -16,6 +16,7 @@
         "mypy_extensions>=0.3",
         "pyblake2>=1.1.1; python_version < '3.6'",
         "requests>=2.20",
+        "retrying==1.3.3",
```
hardbyte: Too specific, suggesting putting `>=1.3`

gusmith:
More changes than expected... Mainly because of the comment:

> wait_exponential_multiplier=100, wait_exponential_max=10000 (with the numbers being configurable)

Changes:
 - create a class for the rest client to contain the server and the retrying configuration. In fact, it makes it cleaner while being able to configure the retrying parameters via the cli. I didn't wanted to use environment variable, because it was not consistent with the current usage of the library
 - retrying configuration uses exponential time growth instead of constant
 - fix a bug introduced by the retrying: when posting the clks, we are reading the data from a file (via `click.File`), so if we are retrying working on the same file, we will read empty value, so we need to reset the reading pointer to the beginning of the file
 - update all the tests to pass
 - reduce requirement restriction from retrying==1.3.3 to retrying>=1.3

gusmith:
@hardbyte all the tests are finally passing (yeah! it was not that simple to run exactly the same tests as the one running on the CI locally).
But just wondering if the provided documentation is enough.

clkhash/rest_client.py
```diff
@@ -210,3 +66,196 @@ def format_run_status(status):
...
+                return
+
+            if old_status != status:
+                yield status
+
+            time.sleep(update_period)
+            old_status = status
+            try:
+                status = self.run_get_status(project, run, apikey)
+            except RateLimitedClient:
```
hardbyte: Probably worth documenting how this interacts with the retries set up inside run_get_status - will this exception ever get here?

tests/test_cli.py
```diff
@@ -14,10 +13,9 @@
 from click.testing import CliRunner
 from future.builtins import range
 
-import clkhash
 import clkhash.cli
 from clkhash import randomnames, rest_client
-from clkhash.rest_client import ServiceError
+from clkhash.rest_client import ServiceError, RestClient
```
hardbyte: The cli has added a few new options - they should at least be tried in a cli test.

tests/test_rest_client.py
```diff
@@ -264,6 +271,9 @@ def test_watch_run_time_out(requests_mock):
 
 
 def test_watch_run_rate_limited(requests_mock):
+    rest_client = RestClient('http://testing-es-url', ClientWaitingConfiguration(wait_exponential_max_ms=10,
```
hardbyte: great to see

clkhash/cli.py
```diff
@@ -100,15 +98,32 @@ def hash(pii_csv, keys, schema, clk_json, quiet, no_header, check_header, valida
 @click.option('--server', type=str, default=DEFAULT_SERVICE_URL, help="Server address including protocol")
 @click.option('-o', '--output', type=click.File('w'), default='-')
 @click.option('-v', '--verbose', default=False, is_flag=True, help="Script is more talkative")
-def status(server, output, verbose):
+@click.option('--client_retrying_wait_exponential_multiplier_ms',
```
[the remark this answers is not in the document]
gusmith: I kept the three options, but simplified their names and their descriptions.
Thanks for the references :)

clkhash/cli.py
```diff
@@ -100,15 +98,32 @@ def hash(pii_csv, keys, schema, clk_json, quiet, no_header, check_header, valida
...
+              default=ClientWaitingConfiguration.DEFAULT_WAIT_EXPONENTIAL_MULTIPLIER_MS,
+              type=int, help="Configuration about retrying a request if the client receives a 503 response."
+                             " See https://pypi.org/project/retrying/ for more documentation.")
```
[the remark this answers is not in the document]
gusmith: :+1:

clkhash/cli.py
```diff
@@ -147,7 +162,20 @@ def status(server, output, verbose):
...
-def create_project(type, schema, server, name, parties, output, verbose):
+@click.option('--client_retrying_wait_exponential_multiplier_ms',
+              default=ClientWaitingConfiguration.DEFAULT_WAIT_EXPONENTIAL_MULTIPLIER_MS,
+              type=int, help="Configuration about retrying a request if the client receives a 503 response."
+                             " See https://pypi.org/project/retrying/ for more documentation.")
+@click.option('--client_retrying_wait_exponential_max_ms',
+              default=ClientWaitingConfiguration.DEFAULT_WAIT_EXPONENTIAL_MAX_MS,
+              type=int, help="Configuration about retrying a request if the client receives a 503 response."
+                             " See https://pypi.org/project/retrying/ for more documentation.")
+@click.option('--client_retrying_stop_max_delay_ms',
```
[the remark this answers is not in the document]
gusmith: Took a bit longer than expected, but the method you provided could not work: all our commands should be based on the same root one `cli` in our code. What we wanted to do was using the same options for multiple commands without creating its own command for that which is slightly more complicated (there was some discussions about it in click, but they won't solve this for some reasons). 

But at the same time, I also modified how we were handling the script verbosity to include it in the command context. So now, 
`clkutil -v status [...]` is equivalent to `clkutil status -v [...]`. Note that some commands are not using verbose, and do not have such options.

clkhash/rest_client.py
```diff
@@ -210,3 +66,196 @@ def format_run_status(status):
...
+        """
+        Return a requests Response object, but will retry the same request as long as the server returns a 503 response and
+        the maximum delayed has not been exceeded.
+        The delay between requests will increase exponentially up to a threshold, from which it stays constant. The delay is:
+        2^x * WAIT_EXPONENTIAL_MULTIPLIER_MS milliseconds between each retry, up to WAIT_EXPONENTIAL_MAX_MS milliseconds,
+        then it stays at WAIT_EXPONENTIAL_MAX_MS milliseconds afterwards."
+        """
+
+        # if client_waiting_configuration is None:
+        #    client_waiting_configuration = ClientWaitingConfiguration()
```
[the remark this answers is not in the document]
gusmith: :+1:

tests/test_cli.py
gusmith: Updated the `help` tests and for all the commands tested which are using the server, I added a test modifying these parameters.

clkhash/rest_client.py
gusmith: Nicely spotted. And it should indeed never been raised. In this case, we could receive a `RetryError` if the retry timeouts.

clkhash/cli.py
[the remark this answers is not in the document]
hardbyte: I'm not understanding why we couldn't have a base for commands that talk to the rest service.
This is much better.

I had a minor refactor where we don't need to rely on storing the verbosity in the context though. See https://github.com/data61/clkhash/compare/fix-272...fix-272-alternative?expand=1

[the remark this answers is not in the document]
gusmith: About your PR: :+1: merged in and it is indeed cleaner.

About the group command, the main issue is that we want to handle options with the group command, but the command type has not been provided yet. So for example, it would look like: `clkutil --server [...] status [$STATUS_OPTIONS]` instead of `clkutil status --server [...] [$STATUS_OPTIONS]`. I'm also not sure how it would behave with the commands not using the server options as the options would need to be provided before the command type... 
This was what I was testing the verbosity: I tried not to include it in the subcommands (e.g. `status`) by only setting it in the click group `cli`. I then realised that I could not do `clkutil status -v` anymore, but only `clkutil -v status` which would have been alright for verbosity, but not nice for setting server attributes.

gusmith:
Note that the builds are failing because our k8s cluster is down. The kube-dns pods are having some issues.

gusmith:
The k8s cluster is back, but the entity service is not deployed...

clkhash/cli.py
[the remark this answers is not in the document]
hardbyte: Ahh ok thanks for the explanation. All good!

gusmith:
The two last commits which have not been reviewed are just:
 - re-running the tutorial python notebooks (I didn't commit the notebook fully ran, so a test fail) without changes in the notebooks
 - merging master branch into this one to be up to date.
I will thus proceed to merge this branch.

Issue #104, opened 2018-04-13
Progress printing should be cli specific


hardbyte:
> Assuming clk.py is meant to be common code that could support a number of
different interfaces, then tqdm’s progress bars (which are specific to a CLI) should be
handled in cli rather than clk - perhaps by passing a progress-bar callback to clk
rather than a “progress_bar” boolean. (Or, could at least reduce code duplication by
having only one call to generate_clks, with callback=None if no progress_bar.)

Aha! Link: https://csiro.aha.io/features/ANONLINK-38

hardbyte:
Part of anonlink-client

PR #278, opened 2019-07-11
Move progress bar from clk to cli

ProgrammedByJake:
For issue #104


clkhash/clk.py
```diff
@@ -101,25 +100,21 @@ def generate_clk_from_csv(input_f,  # type: TextIO
...
-                                    validate=validate,
-                                    callback=callback)
-    else:
-        results = generate_clks(pii_data,
-                                schema,
-                                keys,
-                                validate=validate)
+    progress_callback = None
+    # As we are separating tqdm from the clk module, we don't want to import ProgressBar
+    # We explicitly test for attributes and ignore mypys checks.
```
hardbyte: Interesting approach but it is better to be explicit about what is expected and allowed. So if we define an (optional) type for the Progress Callback we can guarantee that (if it isn't None) there will be an `initialise` method that takes an `int` argument.

The type can be independent of the implementation so you don't have to import tqdm

ProgrammedByJake:
@hardbyte Thanks as always for your guidance. 

I was attempting to find a way to write it with only passing a callback but I had a few issues. 
The main issues were that the cli can't instantiate the progress bar without knowing the size of the data, and the clk performs a print, which if done before closing the progress bar leads to a race condition where you will generally get a progress bar at ~99%, then print statement, then another progress bar at 100%.

My solution was a abstract encoding interface class, allowing the progress encoding interface subclass to be passed from the cli into the clk. The clk then handles the entire lifecycle using a 'with' statement. This cleans up the typing issues and means the clk doesn't have to do any specific calls on the progress bar.

Again, I'm not sure if it's a good approach, but it was cleaner than the few other options I could think of.

hardbyte:
Closing this as abandoned.

PR #280, opened 2019-08-03
Introduce hypothesis to the codebase

hardbyte:
This PR adds the `hypothesis` dependency and adds a small example in testing our string tokenising.


tests/test_tokenizer.py
```diff
@@ -72,6 +79,16 @@ def test_invalid_n(self):
...
 
+    @given(text(min_size=1))
+    def test_string_bigram_token_size(self, myinput):
+        tokens = list(p2_20(myinput))
+        assert len(myinput) == len(tokens) - 1
+
+    @given(text(min_size=1))
+    def test_string_unigram_token_size(self, myinput):
+        tokens = list(p1_20(myinput))
+        assert len(myinput) == len(tokens)
```
wilko77: almost the same code twice.
you could probably generalize this to also include the 'positional' cases.

hardbyte: We might disagree on this, but I don't mind a small amount of code duplication in unit tests. 
I've added a third test for the positional unigram and I'm not fussed if you want to generalize.

Issue #268, opened 2019-07-04
Remove Appveyor and add Azure build status branch protection


gusmith:
Currently, a commit can be merged if it  receives the appveyor build status `continuous-integration/appveyor/pr`.
But we are not replacing appveyor with Azure Build pipeline (merged in master with PR #262), so we should change the branch protection and change the appveyor configuration not to build the project anymore.

hardbyte:
I've added the Azure checks and removed the requirement for appveyor:

![image](https://user-images.githubusercontent.com/855189/62435708-849ba300-b780-11e9-8906-15145ebb9d7a.png)

This issue is therefore to remove the appveyor configuration then?

gusmith:
Hi @MehdiGol 
I can only mentioned what I remember, which is not that much as it happened a few years ago now, but at the time, Azure had really great offers for open source projects to run CI and CD (number of parallel jobs, time to get a worker, etc).
From memory, we found that AppVeyor was quite slow (e.g. https://github.com/data61/clkhash/issues/121) but I do not remember well the details... was it time to get a worker? Was it the worker resources which were a bit small? 
Hoping that help.

PR #281, opened 2019-08-07
Feature full tests azure

gusmith:
Some tests were not ran in the Azure builds, mainly the one using the test entity service.
So in this PR:
 - remove appveyor config file (not required anymore as we are stopping using appveyor, cf issue #268 )
 - run the type checks (as Travis)
 - run the tests requiring the deployed test entity service (at `testing.es.data61.xyz`)
 - run the notebook tests -> actually ignored for windows machine (#283 ) but in the code in case we want to use Azure Dev-Ops for other environment than Windows.


gusmith:
Typecheck tests not working for python 3.5 and 2.7 (tracked in #282 ).
Notebook tests fail on Windows machines (tracked in #283)

.azurePipeline/wholeBuild.yml
```diff
@@ -49,6 +49,19 @@ jobs:
...
+      python -m pip install -U mypy
+      mypy clkhash --ignore-missing-imports --no-implicit-optional --disallow-untyped-calls
+    displayName: 'Typecheck'
+    condition: le(3.6, variables['pythonVersion'])
+  
+  - script: |
+      python -m pip install -U -r docs/doc-requirements.txt
+      pytest --nbval docs -x --sanitize-with docs/tutorial_sanitize.cfg
+    displayName: 'Test notebooks'
+    condition: ne(variables['operatingSystem'], 'vs2017-win2016')
```
hardbyte: Hmm only the notebooks that have anonlink don't work on windows. And as far as I recall we don't actually (yet) build on any non windows pools on Azure DevOps... so this is just here for the future?

gusmith: That's indeed for the future if we decide to move all builds on Azure. I added the choice of operating system for two reasons:
 1) to try it on Windows (and it failed)
 2) the operating system on which to run is already a parameter of the build (cf the file `azurePipeline.yml`) in which we decided to have the operating system environment as a parameter instead of hard coded, which I would assume was for future possibility.

As I added it to test, I found it weird to delete it, which is why a running condition has been added.
But if we are definitely not planning to run this pipeline for otehr OS than Windows, I could remove this step.

.azurePipeline/wholeBuild.yml
```diff
@@ -49,6 +49,19 @@ jobs:
     displayName: 'Test with pytest'
     env:
       INCLUDE_CLI: 1
+      TEST_ENTITY_SERVICE: https://testing.es.data61.xyz
+
+  - script: |
+      python -m pip install -U mypy
+      mypy clkhash --ignore-missing-imports --no-implicit-optional --disallow-untyped-calls
+    displayName: 'Typecheck'
+    condition: le(3.6, variables['pythonVersion'])
```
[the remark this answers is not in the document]
gusmith: Does it need to be executed on different architectures ('x64', 'x86')?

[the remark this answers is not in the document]
hardbyte: Nope, the mypy tool is written in pure python.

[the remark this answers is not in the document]
gusmith: OK, it is run only for `x86` and a selected python version from the main script (currently 3.7).

.azurePipeline/wholeBuild.yml
hardbyte: Okay all good leave the step. 

I'll note that it would be good to run the notebooks that don't depend on anonlink.

.azurePipeline/wholeBuild.yml
```diff
@@ -49,6 +51,34 @@ jobs:
     displayName: 'Test with pytest'
     env:
       INCLUDE_CLI: 1
+      TEST_ENTITY_SERVICE: https://testing.es.data61.xyz
+
+  - script: |
+      python -m pip install -U mypy
+      mypy clkhash --ignore-missing-imports --no-implicit-optional --disallow-untyped-calls
+    displayName: 'Typecheck'
+    condition: and(eq(variables['pythonVersionMypy'], variables['pythonVersion']), eq(variables['architecture'], 'x86'))
```
hardbyte: Out of interest why use a condition? Shouldn't this be in its own job which has a `UsePythonVersion` task? Don't worry about it now, but I'll ask about it for introducing more stages/jobs [here](https://github.com/data61/clkhash/compare/azure-pipeline-stages?expand=1#diff-04fca0223aa9aa49e6a717f8f9cd60cbR23)

docs/devops.rst
```diff
@@ -29,7 +29,9 @@ The build pipeline does:
 
   - install the requirements,
   - package ``clkhash``,
-  - run `pytest`,
+  - run `pytest` (including all the CLI tests and the test requiring a deployed entity service at `https://testing.es.data61.xyz`),
+  - run `mypy` for type checking only a chosen version of Python (currently 3.7)
```
hardbyte:
```suggestion
  - run `mypy` for type checking
```

With mypy all the code is checked

.azurePipeline/wholeBuild.yml
gusmith: I was using a condition because I was expecting that we wanted to run it for each tested OS.
But from your change on your branch, it seems you would like it to be ran only once, so I moved it to its own job (11dca85).

docs/devops.rst
gusmith: :+1:

PR #284, opened 2019-08-08
Azure pipeline refactor

hardbyte:
Just making the azure pipeline job names a bit more human friendly


.azurePipeline/wholeBuild.yml
```diff
@@ -10,15 +10,15 @@ jobs:
       ${{ each py in parameters.pythonVersions }}:
         ${{ each os in parameters.operatingSystems }}:
           ${{ each architecture in parameters.architectures }}:
-            ${{ format('{0}_Python_{1}_{2}', os, py, architecture) }}:
+            ${{ format('Python {0} on {1} ({2})', py, os, architecture) }}:
```
gusmith: Great find: I didn't include spaces because I thought it would be the job name, but it seems that it is only used in the display name and is working :+1:

PR #286, opened 2019-08-09
Feature alternative tokenizer

wilko77:
Can you please check if I'm on the right track with this?

I decided to give `FieldHashingProperties` a new attribute `tokenizer`, which holds the tokenization function. Previously, FHP just had the parameters for tokenization, and then the tokenization function was created on the fly, given the FHP.

Unfortunately, I ran into a minor annoyance with this, as for the process parallelization, the given task has to be pickle-able. This means, that the tokenization function has to be a module level function, and hence the use of functools.partial.

Better ideas?


clkhash/bloomfilter.py
```diff
@@ -287,7 +287,7 @@ def fold_xor(bloomfilter,  # type: bitarray
 
 
 def crypto_bloom_filter(record,      # type: Sequence[Text]
-                        tokenizers,  # type: List[function]
+                        comparators,  # type: List[AbstractComparison]
```
hardbyte: Much nicer

clkhash/comparators.py
```diff
@@ -0,0 +1,117 @@
...
+
+        :param word: The string to tokenize.
+        :return: Iterable of tokens.
+        """
+        pass
+
+
+class NgramComparison(AbstractComparison):
+
+    def __init__(self, n, positional=False):
```
hardbyte: Missing a type annotation for `n`

clkhash/comparators.py
```diff
@@ -0,0 +1,117 @@
...
+        For Example: the 2-grams of 'clkhash' are ' c', 'cl', 'lk', 'kh', 'ha', 'as', 'sh', 'h '. Note the white-
+        space in the first and last token. They serve the purpose to a) indicate the beginning and end of a word, and b)
+        gives every character in the input text a representation in two tokens.
+
+        'n'-gram comparison of strings is tolerant to spelling mistakes, e.g., the strings 'clkhash' and 'clkhush' have
+        6 out of 8 2-grams in common. One wrong character will affect 'n' 'n'-grams. Thus, the larger you choose 'n',
+        the more the error propagates.
+
+        :param n: the n in n-gram, non-negative integer
+        :param positional: enables positional n-gram tokenization
```
hardbyte: The rest of this docstring is 💯 - perhaps worth adding a note about what _positional n-gram tokenization_ actually is

clkhash/comparators.py
```diff
@@ -0,0 +1,117 @@
...
+        if n < 0:
+            raise ValueError('`n` in `n`-gram must be non-negative.')
+        self.n = n
+        self.positional = positional
+
+    def tokenize(self, word):
+        # type: (Text) -> Iterable[Text]
+        """ Produce `n`-grams of `word`.
+
+             :param n: the n in n-gram, non-negative integer
```
hardbyte: nit: these are incorrectly indented

clkhash/comparators.py
```diff
@@ -0,0 +1,117 @@
...
+        This function takes a dictionary, containing the schema definition. It returns a subclass of AbstractComparison.
+    """
+
+    typ = comp_desc.get('type', None)
+
+    if typ == 'ngram':
+        return NgramComparison(comp_desc.get('n', -1), comp_desc.get('positional'))
+    elif typ == 'exact':
+        raise NotImplementedError("I will eventually implement this. Scout's honor.")
+    else:
```
hardbyte: Should there be a way for `get_comparator` to return the null comparator?

wilko77: In my opinion not. 
I doesn't really make sense. It is just there to support some not-so-great legacy code. In fact, this will never actually be called in a clk generation.
The handling of 'ignored' fields should be re-thought instead. The way it is handled in `stream_bloom_filters` and `crypto_bloom_filters` in `bloomfilter.py` is not pretty...

PR #302, opened 2019-10-13
renamed the strategies

### LINKED ISSUE #259: Rename strategy properties in schema v2
> neither "numBits" nor "k" is a good name for a strategy.
> Why don't we name them `fixedK` and `fixedNumBits`

At the same time the properties need to have better descriptions in the schema.

(from https://github.com/data61/clkhash/pull/255#discussion_r292294065)

### LINKED ISSUE #260: Adjust how FieldHashingProperties is constructed
You can instantiate `FieldHashingProperties` with both `k` and `num_bits`.

Strategy should be its own thing that's either fixed_k or (exclusive) fixed_num_bits.

wilko77:
In the days of old, we had as strategies:
- k 
- numBits

My issue with this is, that it doesn't describe a strategy, this are names of variables. Also, for the uninitiated, it is very hard to guess what these actually do.

Therefore, for a brighter future, I present the new and shiny strategy names:
- bitsPerToken
- bitsPerFeature

 We had some other contestants. There were also `fixed_k`, `variable_k`, and `fixed_numBits`. I decided against all of those, as they use very domain specific names ('k' and 'numBits') which someone, who hasn't got a deep understanding of our library or the corresponding papers, cannot understand.

(closes #259)

Main contributions:
1) disambiguate the strategy in FieldHashingProperties.
Previously, one could provide both, `numBits` and `k` as kw-arguments to FieldHashingProperties. Thus, it appears that they are optional, or that you can set both. 
Now you have to provide a non-optional strategy argument which has to be of type `StrategySpec`.
In the spirit of the `field_formats.py` file, this is a meta class with two implementations: `BitsPerTokenStrategy` and `BitsPerFeatureStrategy`. (fixes #260)

2) explicit v2 to v3 conversion.
Just as the already existing v1 to v2 conversion function, I added a v2 to v3 version of this.

3) adjusted all tests, examples and docs to the new strategies.


gusmith:
First, I like the new strategy names :+1: 
Second, more of a question: would/could we have strategies requiring more than a single parameter?
Currently, the json would look like:
```
strategy: {
  bitsPerToken: 100
}
```
but with attributes, it would look like:
```
strategy: {
  strategy_which_has_attributes: {
    attribute_1: foo1,
    attribute_2: foo2
  }
}
```
So I'm wondering, should we add a `type` key in the `strategy` object, describing the type of strategy, and then providing the parameters?
```
strategy: {
  type: strategy_which_has_attributes,
  attribute_1: foo1,
  attribute_2: foo2
}
```
or 
```
```
strategy: {
  type: bitsPerToken,
  k: 100
}
```

**Comment by gusmith** [comment]
I am removing myself from the reviewer, to get notified when you would like a new review. I'm still watching the PR, so I get notifications on comments too.

File: clkhash/field_formats.py [hunk 1]
```diff
@@ -69,6 +68,76 @@ def from_json_dict(cls, json_dict):
         )
 
 
+@add_metaclass(abc.ABCMeta)
+class StrategySpec(object):
+    """ Stores the information about the insertion strategy.
+
+    A strategy has to implement the 'ks' function, which defines how often a token gets inserted into the Bloom filter.
```
hardbyte: `ks` could do with a more descriptive name.

clkhash/field_formats.py
```diff
@@ -69,6 +68,76 @@ def from_json_dict(cls, json_dict):
...
+    """
+
+    @abc.abstractmethod
+    def ks(self, num_tokens):
+        # type: (int) -> List[int]
+        """ Return a list of 'num_tokens' 'k' values, defining how often a token gets inserted into the Bloom filter.
```
hardbyte: Can you reword this. I had trouble understanding it until I read the concrete classes below.

clkhash/field_formats.py
```diff
@@ -69,6 +68,76 @@ def from_json_dict(cls, json_dict):
...
+        else:
+            raise InvalidSchemaError('unknown strategy')
+
+
+class BitsPerTokenStrategy(StrategySpec):
+    """ Insert every token the same number of times.
+
+    This is the strategy from the original Schnell paper. The provided value 'bits_per_token' (the 'k' value in the
+    paper) defines the number of hash functions that are used to insert each token into the Bloom filter.
+
```
hardbyte: Maybe worth adding a note here about the impact/risk of unequal sized features.

clkhash/field_formats.py
```diff
@@ -69,6 +68,76 @@ def from_json_dict(cls, json_dict):
...
+                 ):
+        # type: (...) -> None
+        self.bits_per_token = bits_per_token
+
+    def ks(self, num_tokens):
+        # type: (int) -> List[int]
+        return [self.bits_per_token] * num_tokens
+
+
+class BitsPerFeatureStrategy(StrategySpec):
```
hardbyte: Perhaps worth renaming a few other places to use `feature` rather than `field`.
E.g. `FieldHashingProperties`, this file.

clkhash/schema.py
```diff
@@ -147,6 +147,33 @@ def convert_feature(f):
     return result
 
 
+def convert_v2_to_v3(
```
hardbyte: 👍

clkhash/field_formats.py
wilko77: out of scope. I created an issue for that.

PR #309, opened 2019-10-31
Feature cli schema converter

### LINKED ISSUE #206: cli to convert v1 to latest schema json file.
Add cli command to convert v1 schema json file to vX schema json file.

Use same conversion, schema.convert_v1_to_v2,  that is applied when a v1 schema file is loaded.

wilko77:
I added a new command for the cli to convert a schema file to the latest version.

closes #206

### DIFF
--- a/clkhash/schema.py
+++ b/clkhash/schema.py
@@ -94,10 +94,8 @@ class Schema:
         return "<Schema (v3): {} fields>".format(len(self.fields))
 
 
-def _convert_v1_to_v2(
-        dict  # type: Dict[str, Any]
-    ):
-    # type: (...) -> Dict[str, Any]
+def _convert_v1_to_v2(dict):
+    # type: (Dict[str, Any]) -> Dict[str, Any]
     """
     Convert v1 schema dict to v2 schema dict.
     :param dict: v1 schema dict
@@ -147,10 +145,8 @@ def _convert_v1_to_v2(
     return result
 
 
-def _convert_v2_to_v3(
-        dict  # type: Dict[str, Any]
-    ):
-    # type: (...) -> Dict[str, Any]
+def _convert_v2_to_v3(dict):
+    # type: (Dict[str, Any]) -> Dict[str, Any]
     """
     Convert v2 schema dict to v3 schema dict.
     :param dict: v2 schema dict
@@ -175,6 +171,7 @@ def _convert_v2_to_v3(
 
 
 def convert_to_latest_version(schema_dict):
+    # type: (Dict[str, Any]) -> Dict[str, Any]
     """ Convert the given schema to latest schema version.
 
      :param schema_dict: A dictionary describing a linkage schema. This dictionary must have a `'features'` key
--- a/clkhash/schema.py
+++ b/clkhash/schema.py
@@ -174,9 +174,9 @@ def convert_to_latest_version(schema_dict):
     # type: (Dict[str, Any]) -> Dict[str, Any]
     """ Convert the given schema to latest schema version.
 
-     :param schema_dict: A dictionary describing a linkage schema. This dictionary must have a `'features'` key
-            specifying the columns of the dataset. It must have a `'version'` key containing the master schema version
-            that this schema conforms to. It must have a `'hash'` key with all the globals.
+     :param schema_dict: A dictionary describing a linkage schema. This dictionary must have a `'version'` key
+            containing a master schema version. The rest of the schema dict must conform to the corresponding
+            master schema.
      :return: schema dict of the latest version
      raises SchemaError if schema version is not supported"""
     version = schema_dict.get('version', "'not specified'")
--- a/clkhash/schema.py
+++ b/clkhash/schema.py
@@ -94,19 +94,19 @@ class Schema:
         return "<Schema (v3): {} fields>".format(len(self.fields))
 
 
-def _convert_v1_to_v2(dict):
+def _convert_v1_to_v2(schema_dict):
     # type: (Dict[str, Any]) -> Dict[str, Any]
     """
     Convert v1 schema dict to v2 schema dict.
-    :param dict: v1 schema dict
+    :param schema_dict: v1 schema dict
     :return: v2 schema dict
     """
-    dict = deepcopy(dict)
-    version = dict['version']
+    schema_dict = deepcopy(schema_dict)
+    version = schema_dict['version']
     if version != 1:
         raise ValueError('Version {} not 1'.format(version))
 
-    clk_config = dict['clkConfig']
+    clk_config = schema_dict['clkConfig']
     k = clk_config.pop('k')
     clk_hash = clk_config['hash']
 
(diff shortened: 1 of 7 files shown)


clkhash/schema.py
```diff
@@ -94,10 +94,8 @@ def __repr__(self):
         return "<Schema (v3): {} fields>".format(len(self.fields))
 
 
-def convert_v1_to_v2(
-        dict  # type: Dict[str, Any]
-    ):
-    # type: (...) -> Dict[str, Any]
+def _convert_v1_to_v2(dict):
```
gusmith: We shouldn't be using the keyword `dict` for a parameter.

clkhash/schema.py
```diff
@@ -174,6 +170,27 @@ def convert_v2_to_v3(
...
+            that this schema conforms to. It must have a `'hash'` key with all the globals.
+     :return: schema dict of the latest version
+     raises SchemaError if schema version is not supported"""
+    version = schema_dict.get('version', "'not specified'")
+    if version not in MASTER_SCHEMA_FILE_NAMES.keys():
+        msg = ('Schema version {} is not supported. '
+               'Consider updating clkhash.').format(version)
+        raise SchemaError(msg)
+    if schema_dict['version'] == 1:
+        schema_dict = _convert_v1_to_v2(schema_dict)
```
gusmith: Shouldn't we also validate that the output schema when modified?

clkhash/schema.py
```diff
@@ -174,6 +170,27 @@ def convert_v2_to_v3(
...
+            specifying the columns of the dataset. It must have a `'version'` key containing the master schema version
+            that this schema conforms to. It must have a `'hash'` key with all the globals.
+     :return: schema dict of the latest version
+     raises SchemaError if schema version is not supported"""
+    version = schema_dict.get('version', "'not specified'")
+    if version not in MASTER_SCHEMA_FILE_NAMES.keys():
+        msg = ('Schema version {} is not supported. '
+               'Consider updating clkhash.').format(version)
+        raise SchemaError(msg)
+    if schema_dict['version'] == 1:
```
gusmith: Could be nice to have a loop instead of doing if statements. Just thinking for the future, when we will have more schemas, not to have to modify again this method.

clkhash/schema.py
```diff
@@ -174,6 +170,27 @@ def convert_v2_to_v3(
...
+        schema_dict = _convert_v1_to_v2(schema_dict)
+    if schema_dict['version'] == 2:
+        schema_dict = _convert_v2_to_v3(schema_dict)
+    return schema_dict
+
+
 def from_json_dict(dct, validate=True):
     # type: (Dict[str, Any], bool) -> Schema
     """ Create a Schema for v1 or v2 according to dct
```
gusmith: Not for v3?

clkhash/schema.py
```diff
@@ -174,6 +170,27 @@ def convert_v2_to_v3(
     return dict
 
 
+def convert_to_latest_version(schema_dict):
+    # type: (Dict[str, Any]) -> Dict[str, Any]
+    """ Convert the given schema to latest schema version.
+
+     :param schema_dict: A dictionary describing a linkage schema. This dictionary must have a `'features'` key
```
gusmith: Shouldn't we simply say that it should have a version key, and follow the corresponding schema as described in the file `vx.json`?

clkhash/schema.py
```diff
@@ -192,19 +209,9 @@ def from_json_dict(dct, validate=True):
...
-        if validate:
-            validate_schema_dict(dct)
-    elif version not in MASTER_SCHEMA_FILE_NAMES.keys():
-        msg = ('Schema version {} is not supported. '
-               'Consider updating clkhash.').format(version)
-        raise SchemaError(msg)
-
+    dct = convert_to_latest_version(dct)
+    if validate:
+        validate_schema_dict(dct)
```
gusmith: Just wondering if the validation shouldn't be done after creation, all the time. It seems weird to me to return a transformed schema which will not be valid...

tests/test_cli.py
```diff
@@ -129,6 +130,8 @@ def test_bench(self):
         assert 'hashes in' in result.output
 
 
+@unittest.skipUnless("INCLUDE_CLI" in os.environ,
```
gusmith: :+1:

tests/test_cli.py
```diff
@@ -163,6 +166,50 @@ def test_bad_v2_schema(self):
         assert 'schema is not valid.' in result.output
 
 
+@unittest.skipUnless("INCLUDE_CLI" in os.environ,
```
gusmith: In the class `TestSchemaConversionCommand` and `TestSchemaValidationCommand`, there are no tests for the schema v3. As it is not really part of this PR, we should at least open a issue for it, and solve that soon as possible.

tests/test_schema.py
```diff
@@ -84,11 +84,22 @@ def test_missing_master(self):
...
+        schema_v3 = schema._convert_v2_to_v3(schema_v2)
+        assert schema_v3['version'] == 3
+        schema.validate_schema_dict(schema_v3)
+
+    def test_convert_schema_to_latest(self):
+        schema_v1 = _schema_dict(DATA_DIRECTORY, 'randomnames-schema.json')
+        assert schema_v1['version'] == 1
+        newest_schema = schema.convert_to_latest_version(schema_v1)
+        schema.validate_schema_dict(newest_schema)
+        assert newest_schema['version'] == max(schema.MASTER_SCHEMA_FILE_NAMES.keys())
```
gusmith: Ready for future schema versions :+1:

tests/test_schema.py
```diff
@@ -84,11 +84,22 @@ def test_missing_master(self):
 
         schema.MASTER_SCHEMA_FILE_NAMES = original_paths
 
-    def test_convert_v1_to_v2(self):
+    def test_schema_conversion(self):
         schema_v1 = _schema_dict(DATA_DIRECTORY, 'randomnames-schema.json')
```
gusmith: Just looking at this file `randomnames-schema.json`, there are some weird things:
 - the file is used for tests but in the main library data folder (which could be OK)
 - the main library (from the cli methods `generate` and `generate_default_schema`) is creating data which is not using our latest schema version... 
that's not part of this PR, but we should track this.

tests/test_schema.py
```diff
@@ -84,11 +84,22 @@ def test_missing_master(self):
...
+        assert schema_v1['version'] == 1
+        schema_v2 = schema._convert_v1_to_v2(schema_v1)
+        assert schema_v2['version'] == 2
         schema.validate_schema_dict(schema_v2)
+        schema_v3 = schema._convert_v2_to_v3(schema_v2)
+        assert schema_v3['version'] == 3
+        schema.validate_schema_dict(schema_v3)
+
+    def test_convert_schema_to_latest(self):
+        schema_v1 = _schema_dict(DATA_DIRECTORY, 'randomnames-schema.json')
```
gusmith: Same comment as in line 88.

clkhash/schema.py
wilko77: `from_json_dict` already does that. I don't want to validate too many times.
In the CLI, there is a separate command for validation

tests/test_schema.py
wilko77: -> #310

clkhash/schema.py
gusmith: But `from_json_dict` is not called from the cli, so currently running the `convert-schema` command does not validate the returned result.

wilko77:
sorry, forgot about the docs...

PR #313, opened 2019-11-01
named coverage reports

wilko77:
we were pushing nameless reports to codecov, which makes it very hard, to identify a specific coverage report.
I change that:
- we send the environment variables TRAVIS_OS_NAME,TRAVIS_PYTHON_VERSION,INCLUDE_CLI
- we also send a name as "$TRAVIS_OS_NAME"_"$TRAVIS_PYTHON_VERSION""$EXTRAS"

On the side, codecov is sometimes prematurely reporting coverage back to github which then wrongly claims that the coverage reduced by a large amount. I added a section that codecov has to wait for at least 5 reports before reporting. (No idea if that's enough...)


wilko77:
I massaged azure as well. However, the codecov integration into azure doesn't work quite as well as with Travis. On azure it assigns the report to a different commit than on travis.

hardbyte:
Doesn't codecov already publish the travis build information (just not as environment variables)? We have to manually do it in Azure because codecov.io doesn't natively support azure right?

image

wilko77:
It doesn't give good names though. With proper naming you can do things like this:
image

gusmith:
It's actually really weird: as @wilko77 mentioned, Azure was sometime pushing to the wrong commit, but on the latest commit, Azure did the what it is supposed to, but Travis didn't... Could codecov accept reports from only one source??? That would be weird...

PR #316, opened 2019-11-05
Feature numeric comparison

### LINKED ISSUE #299: implement fuzzy integer comparison
Clkhash now has built-in support for different comparison methods. To expand our offering in that space, I propose to implement fuzzy integer comparison.

Use the idea of BitVectors (see #3) to implement integer comparison where the difference between two integer defines the similarity.
Follow the structure of #285 and take #290 as an example.

wilko77:
This is an implementation of the idea from Vatsalan and Christen (Privacy-preserving matching of similar patients, Journal of Biomedical Informatics, 2015).
It creates tokens from numbers, such that numbers that a closer together have more tokens in common.
It works with both, integer and floats. 

My first implementation was using the decimal module, which was quite nice. However, the test ran for too long, as the numbers could get quite big and then the hypothesis tests would time out.
Now I am using floats. This introduces some floating point problems. For instance, there is that arbitrary jump from 1.something_e308 to infinity. 
That's why we don't support the full range of floats any more. 

closes #299


clkhash/comparators.py
```diff
@@ -93,6 +93,87 @@ def tokenize(self, word):  # type: (Text) -> Iterable[Text]
...
+            raise ValueError('not enough fractional precision to encode threshold_distance')
+        self.resolution = resolution
+        self.fractional_precision = fractional_precision
+        # instead of dividing threshold distance as in the paper, we rather multiply the inputs by 'resolution' and then
+        # use threshold_distance as distance_interval (saves a division which would need more precision)
+        self.distance_interval = self.threshold_distance
+
+    def tokenize(self, word):  # type: (Text) -> Iterable[Text]
+        try:
+            v = int(word, base=10)  # we try int first, so we don't loose precision
```
gusmith: But it still must be multiplied by `pow(10, self.fractional_precision)`, in fact we could have the values `3` and `3.0`, which should return the same tokens. Currently it does not:
```
def testouille():
    comp = NumericComparison(threshold_distance=5, resolution=3, fractional_precision=2)
    cand_tokens = comp.tokenize("3")
    other_tokens = comp.tokenize("3.0")
    other_tokens2 = comp.tokenize("3.1")
    print("{}\n\n{}\n\n{}".format(cand_tokens, other_tokens, other_tokens2))
```
returns:
```
['-1500', '-1000', '-500', '0', '500', '1000', '1500']

['500', '1000', '1500', '2000', '2500', '3000', '3500']

['500', '1000', '1500', '2000', '2500', '3000', '3500']
```

clkhash/comparators.py
```diff
@@ -93,6 +93,87 @@ def tokenize(self, word):  # type: (Text) -> Iterable[Text]
...
+                          are further apart have no elements in common. (*)
+    - resolution: controls how many tokens are generated. (the 'b' in the paper). Given an interval of size
+                  'threshold_distance' we create 'resolution tokens to either side of the mid-point plus one token for
+                  the mid-point. Thus,  2 * resolution + 1 tokens in total. A higher resolution differentiates better
+                  between different values, but should be chosen such that it plays nicely with the overall Bloom filter
+                  size and insertion strategy.
+
+    (*) the reality is a bit more tricky. Depending on the choice of parameters we first have to quantize the inputs, in
+    order to get comparable tokens. For example, if we choose a 'threshold_distance' of 8 and a 'resolution' of 2, then
+    the tokens for x=25 would be [21, 23, 25, 27, 29] and for y=26 [22, 24, 26, 28, 30], resulting in sets with no
```
gusmith: With the provided parameters, the tokens are actually `['88', '96', '104', '112', '120']` for `x=25`, `threshold_distance=8` and a `resolution=2`.

clkhash/schemas/v3.json
```diff
@@ -157,7 +157,7 @@
           {
             "required": ["bitsPerToken"],
             "bitsPerToken": {
-              "type": "number",
+              "type": "integer",
```
gusmith: :+1:

tests/test_comparators.py
```diff
@@ -85,6 +88,139 @@ def test_exact_num_tokens(word):
...
+@given(thresh_dist=floats(min_value=0.0, allow_nan=False, allow_infinity=False, max_value=1e287),
+       resolution=integers(min_value=1, max_value=256),
+       precision=integers(min_value=0, max_value=20),
+       candidate=floats(allow_nan=False, allow_infinity=False, max_value=1e288, min_value=-1e287))
+def test_numeric_properties(thresh_dist, resolution, precision, candidate):
+    assume(thresh_dist > 0)
+    adj_dist = thresh_dist * pow(10, precision)
+    if int(round(adj_dist)) <= 0:
+        with pytest.raises(ValueError):
+            NumericComparison(thresh_dist, resolution, precision)
```
gusmith: Just wondering, should we split this in two independent tests? One assuming that `int(round(adj_dist)) <= 0` and the other assuming the other condition? 
Actually, the second condition could become a default method to be ran on ever returned tokens, also used in the two following tests.

tests/test_comparators.py
```diff
@@ -85,6 +88,139 @@ def test_exact_num_tokens(word):
...
+    else:
+        numbers = [candidate + int(thresh_dist * (i * 0.1)) for i in range(20)]
+    def overlap(other):
+        other_tokens = comp.tokenize(str(other))
+        return len(set(cand_tokens).intersection(set(other_tokens)))
+    overlaps = [overlap(num) for num in numbers]
+    assert overlaps[0] == len(cand_tokens)
+    assert overlaps[-1] == 0
+    assert all(x >= y for x, y in zip(overlaps, overlaps[1:])), 'with increasing distance, the overlap reduces'
+
```
gusmith: Missing test with mix of integers and floats, leading to the problem described in another comment in the class.

tests/testdata/good-schema-v3.json
```diff
@@ -193,11 +193,11 @@
...
         "comparison": {
-          "type": "ngram",
-          "n": 1,
-          "positional": true
+          "type": "numeric",
+          "thresholdDistance": 100,
+          "resolution": 50
         },
-        "strategy": {"bitsPerFeature":  100}
+        "strategy": {"bitsPerFeature":  101}
```
gusmith: I'm curious, why did you increased it by 1?

gusmith:
I also forgot to mention, but the documentation needs to be updated too (example `schema.rst`), and maybe some tutorials...

clkhash/comparators.py
```diff
@@ -93,6 +93,87 @@ def tokenize(self, word):  # type: (Text) -> Iterable[Text]
...
+    would be quantized to even numbers. Thus x=25 would be mapped to 26, and z=24.99 would be mapped to 24.
+    The quantization has the side effect that sometimes two values who are further than 'threshold_distance' but not
+    more than 'threshold_distance' + 1/2 quantization level apart can share a common token.
+
+    We are dealing with floating point numbers by quantising them to integers by multiplying them with
+    10 ** 'fractional_precision' and then rounding them to the nearest integer.
+
+    Thus, we don't support to full range of floats, but the subset between
+    2.2250738585072014e-(308 - fractional_precision - log(resolution, 10)) and
+    1.7976931348623157e+(308 - fractional_precision - log(resolution, 10))
```
hardbyte: 😮

clkhash/comparators.py
```diff
@@ -93,6 +93,87 @@ def tokenize(self, word):  # type: (Text) -> Iterable[Text]
...
+
+    (*) the reality is a bit more tricky. Depending on the choice of parameters we first have to quantize the inputs, in
+    order to get comparable tokens. For example, if we choose a 'threshold_distance' of 8 and a 'resolution' of 2, then
+    the tokens for x=25 would be [21, 23, 25, 27, 29] and for y=26 [22, 24, 26, 28, 30], resulting in sets with no
+    common element. The quantization ensures that the inputs are mapped onto a common grid. In our example, the values
+    would be quantized to even numbers. Thus x=25 would be mapped to 26, and z=24.99 would be mapped to 24.
+    The quantization has the side effect that sometimes two values who are further than 'threshold_distance' but not
+    more than 'threshold_distance' + 1/2 quantization level apart can share a common token.
+
+    We are dealing with floating point numbers by quantising them to integers by multiplying them with
```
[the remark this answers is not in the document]
wilko77: that's the british spelling. _z_ is american.

clkhash/comparators.py
wilko77: good point, I have not though about the use case where the inputs are a mixture of floats and integers.

tests/testdata/good-schema-v3.json
wilko77: the numeric comparison produces 2 * resolution + 1 = 101 tokens.

clkhash/comparators.py
wilko77: I kept the example in line with the paper. The actual tokens are more an implementation detail which might produce more confusion than clarification.

tests/test_comparators.py
wilko77: `adj_dist` depends on two inputs: `thresh_dist` and `precision`. In order to separate the tests you would have to define a custom input generation strategy which assures that the combination of those two is always valid. 
I don't think this is worth the effort.

docs/schema.rst
```diff
@@ -347,6 +348,21 @@ type         string                  no       has to be 'exact'
 ===========  =====================   ======== ===========
 
 
+.. _schema/numericComparison:
+
+numeric comparison
+^^^^^^^^^^^^^^^^^^
+
+Numerical comparisons of integers or floating point numbers such that the distance between two numbers relate to the similarity of the produced tokens. Also see the `API docs for NgramComparison <clkhash.html#clkhash.comparators.NumericComparison>`_
```
gusmith: I think that's a copy/paste issue: `API docs for NgramComparison` should be `API docs for NumericComparison`?

clkhash/comparators.py
```diff
@@ -114,15 +114,16 @@ class NumericComparison(AbstractComparison):
...
-    would be quantized to even numbers. Thus x=25 would be mapped to 26, and z=24.99 would be mapped to 24.
-    The quantization has the side effect that sometimes two values who are further than 'threshold_distance' but not
-    more than 'threshold_distance' + 1/2 quantization level apart can share a common token.
-
-    We are dealing with floating point numbers by quantising them to integers by multiplying them with
+    (*) the reality is a bit more tricky. We first have to quantize the inputs to multiples of 'threshold_distance' /
+    (2 * resolution), in order to get comparable tokens.
+    For example, if we choose a 'threshold_distance' of 8 and a 'resolution' of 2, then the tokens for x=25 would be
+    [21, 23, 25, 27, 29] and for y=26 [22, 24, 26, 28, 30], resulting in sets with no common element. The quantization
+    ensures that the inputs are mapped onto a common grid. In our example, the values would be quantized to even
```
gusmith: I'm still surprised to see an example in the documentation which the code is actually not returning. I understand that is what the paper would return, but as we have an "inspired" implementation, our tokens are equivalent but in a different scale.

PR #319, opened 2019-11-14
Moving all the travis steps into azure devops

gusmith:
In this PR, we are planning to decommission travis to build clkhash.
So we will move all the tests into azure Devops.
When the PR will be merged, we will need to remove the requirement on the travis checks, which I cannot do (I do not have the access in github to the branch protection settings).


.azurePipeline/wholeBuild.yml
```diff
@@ -2,19 +2,23 @@ parameters:
   pythonVersions: []
   operatingSystems: []
   architectures: []
+  fullTests: False
 
 jobs:
-- job: 'test_and_publish'
```
gusmith: The name of the job has been removed to use a random one as we are now using this template multiple times in the same build (for different use cases). But the user "pretty" view in all cases do not see this name, she only sees the display name, so it is staying as pretty as it was.

.azurePipeline/wholeBuild.yml
```diff
@@ -2,19 +2,23 @@ parameters:
...
+- job:
   strategy:
-    maxParallel: 2
+    ${{ if eq(parameters.fullTests, True) }}:
+      maxParallel: 2
```
gusmith: We were reducing the parallelism of the builds because of the deployed entity-service tests, so now we do not limit it if we run local tests (i.e. without the CLI and without the deployed service).

.azurePipeline/wholeBuild.yml
```diff
@@ -2,19 +2,23 @@ parameters:
...
     matrix:
       ${{ each py in parameters.pythonVersions }}:
         ${{ each os in parameters.operatingSystems }}:
           ${{ each architecture in parameters.architectures }}:
-            ${{ format('Python {0} on {1} ({2})', py, os, architecture) }}:
+            ${{ format('Python {0} on {1} ({2}), full tests {3}', py, os, architecture, parameters.fullTests) }}:
```
gusmith: I am open to suggestion for more descriptive names

.azurePipeline/wholeBuild.yml
```diff
@@ -95,9 +101,14 @@ jobs:
       failIfCoverageEmpty: true
 
   - bash: |
+        report_name="$(operatingSystem)_$(architecture)_$(pythonVersion)"
+        if [ $(fullTests) = "True" ]
+        then
+          report_name=${report_name}_full_tests
+        fi
```
gusmith: Include in the report name the fact that all the tests have been ran or not.

gusmith:
Before merging in, I will delete the Travis file, but I let it in for the moment for the reviewer to check that what is tested in travis is also in the azure pipeline.

.azurePipeline/wholeBuild.yml
```diff
@@ -50,31 +56,18 @@ jobs:
     displayName: 'Test with pytest'
     timeoutInMinutes: 10 # it should pass in 2.5 minutes, so 10 should be more than enough even if the server is busy.
     env:
-      INCLUDE_CLI: 1
-      TEST_ENTITY_SERVICE: https://testing.es.data61.xyz
+      ${{ if eq(parameters.fullTests, True) }}:
+        INCLUDE_CLI: 1
+        TEST_ENTITY_SERVICE: https://testing.es.data61.xyz
```
hardbyte: I think it would be worth exposing this URL as an azure variable? That way we can use the clkhash tests to test a different deployment of the entity service (e.g. a PR branch or alpha release)

.azurePipeline/wholeBuild.yml
```diff
@@ -50,31 +56,18 @@ jobs:
...
-  # The list of tutorials without anonlink are provided in the file `docs/list_tutorials_without_anonlink.txt`.
-  # The anonlink requirement is in its own dependency file `docs/doc-requirements-anonlink.txt` while all
-  # the other required dependencies are in `docs/doc-requirements-others.txt`.
-  # Note also that with the current requirements,
+# This step is currently not working on windows operating system, because anonlink must be compiled to be installed.
+# The change will come with the future release 0.13. Tracked via issue #321
   - script: |
-      python -m pip install -U -r docs/doc-requirements-anonlink.txt -r docs/doc-requirements-others.txt
+      pip install -e .
+      python -m pip install -U -r docs/doc-requirements.txt
```
hardbyte: Nice, I'm glad we could simplify that!

.azurePipeline/wholeBuild.yml
```diff
@@ -50,31 +56,18 @@ jobs:
...
-  #    python -m pip install -U -r docs/doc-requirements-others.txt
-  #    list_file=''
-  #    for file in $(cat docs/list_tutorials_without_anonlink.txt)
-  #    do
-  #      list_file="$list_file docs/$file"
-  #    done
-  #    pytest --nbval -x --sanitize-with docs/tutorial_sanitize.cfg $list_file
-  #  displayName: 'Test notebooks without anonlink'
-  #  condition: and(eq(variables['operatingSystem'], 'vs2017-win2016'), ne(variables['pythonVersion'], '2.7'))
+    condition: and(ne(variables['pythonVersion'], '2.7'), eq(variables['fullTests'], True), ne(variables['operatingSystem'], 'vs2017-win2016'))
```
hardbyte: Is it worth a comment saying we don't test notebooks on Python 2.7 (presumably because anonlink is only for Python3+?)

azurePipeline.yml
```diff
@@ -45,12 +45,38 @@ stages:
...
+      operatingSystems: ['vs2017-win2016']
+      architectures: ['x64']
+  - template: .azurePipeline/wholeBuild.yml  # Template reference
+    parameters:
+      pythonVersions: ['3.8', '3.6', '3.5', '2.7']
+      operatingSystems: ['macos-10.13']
+      architectures: ['x64']
+  - template: .azurePipeline/wholeBuild.yml  # Template reference
+    parameters:
+      pythonVersions: ['3.6', '3.5', 'pypy3', 'pypy2']
```
hardbyte: Good coverage!

docs/devops.rst
```diff
@@ -42,6 +41,39 @@ The build pipeline requires one environment variable provided by Azure environme
 
 Most of the complexity is abstracted into the template in `.azurePipeline/wholeBuild.yml`.
 
+Description of what is tested:
+
+==================   ====================  ===============  ===========  ==================  =========
+Python Version       Operating System      Standard pytest  INLCUDE_CLI  ENTITY_SERVICE_URL  Notebooks
```
hardbyte: 💯

.azurePipeline/wholeBuild.yml
```diff
@@ -95,9 +88,14 @@ jobs:
       failIfCoverageEmpty: true
 
   - bash: |
+        report_name="$(operatingSystem)_$(architecture)_$(pythonVersion)"
+        if [ $(fullTests) = "True" ]
```
wilko77: is it really just one '=' in bash for equality check?

.azurePipeline/wholeBuild.yml
gusmith: I'll configure that to happen, but there is one issue with this solution: the github checks.
Let me explain: a run on Azure is associated to a commit. When the run progress, it sends the status of the build to github for the associated commit.
Now, if we just want to test something "for fun", I would assume we do not want to send such status checks (or at least, have a different name). The main reason for not sending this checks is not to block a PR, or not to show that `master` (or branch `foo`) is broken. 
In my mind, the status reported in github should be in the standard configuration, not for testing extra things.
I'm currently looking to see if it is possible to deactivate the status updates from the yaml description of the pipeline, but that seems quite tricky.

Otherwise, we could add a new pipeline which wouldn't report any status, or different status which are not required by github branch protection.

.azurePipeline/wholeBuild.yml
gusmith: It has been simplified because in all cases we were not running these tests on Windows. At least we tried :)

.azurePipeline/wholeBuild.yml
gusmith: :+1:

.azurePipeline/wholeBuild.yml
```diff
@@ -2,19 +2,25 @@ parameters:
...
     matrix:
       ${{ each py in parameters.pythonVersions }}:
         ${{ each os in parameters.operatingSystems }}:
           ${{ each architecture in parameters.architectures }}:
-            ${{ format('Python {0} on {1} ({2})', py, os, architecture) }}:
+            ${{ format('Python {0} on {1} ({2}), full tests {3}', py, os, architecture, parameters.fullTests) }}:
               pythonVersion: ${{ py }}
               operatingSystem: ${{ os }}
               architecture: ${{ architecture }}
+              fullTests: ${{ parameters.FullTests }}
```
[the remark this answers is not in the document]
gusmith: Nicely spotted.

.azurePipeline/wholeBuild.yml
gusmith: It works with one (from http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html):

> "=" may be used instead of "==" for strict POSIX compliance.

but cleaner with two indeed. So updated.

.azurePipeline/wholeBuild.yml
gusmith: I have not find any configuration allowing this kind of behavior. So you can now start a manual run and provide a new url for the entity-service, but use at your own risk.

Issue #339, opened 2020-02-28
Use of ProcessPoolExecutor prevents use in Lambda functions


doc-E-brown:
AWS Lambda functions does not support the use of multiprocessing and thus prevents the use of generate_clk within Lambda functions.   The specific line causing the issue is

https://github.com/data61/clkhash/blob/a3f67e9e19e478ce29c06021aed91ce92df8494b/clkhash/clk.py#L157

While a significant performance hit would occur, allowing the use of ThreadPoolExecutor would allow use in Lambda functions without changing any of the code execution steps.  Having a flag within the function signature which defaults to the existing ProcessPoolExecutor but allows ThreadPoolExecutor would allow use within AWS Lambda functions without changing the existing API.

PR #340, opened 2020-02-28
Allow ThreadPoolExecutor for use in AWS Lambda functions

doc-E-brown:
This PR addresses #339 in providing a flag within the generate_clk function signature to use a ThreadPoolExecutor as opposed to ProcessPoolExecutor.  This change is propagated throughout the function signatures that implement generate_clk as well as the cli and benchmark cli. 

Documentation has also been updated to accommodate the change.

### DIFF
--- a/clkhash/clk.py
+++ b/clkhash/clk.py
@@ -2,11 +2,11 @@
 Generate CLK from data.
 """
 
-import concurrent.futures
+from concurrent.futures import (ProcessPoolExecutor, ThreadPoolExecutor)
 import logging
 import time
 from typing import (AnyStr, Callable, cast, Iterable, List, Optional,
-                    Sequence, TextIO, Tuple, TypeVar, Union)
+                    Sequence, TextIO, Tuple, TypeVar, Union, Type)
 from future.builtins import range
 from tqdm import tqdm
 
@@ -165,11 +165,8 @@ def generate_clks(pii_data,  # type: Sequence[Sequence[str]]
     futures = []
 
     # Compute Bloom filter from the chunks and then serialise it
-
-    if use_multiprocessing:
-        pool_executor = concurrent.futures.ProcessPoolExecutor
-    else:
-        pool_executor = concurrent.futures.ThreadPoolExecutor
+    pool_executor = ProcessPoolExecutor if use_multiprocessing else \
+        ThreadPoolExecutor # type: Union[Type[ProcessPoolExecutor], Type[ThreadPoolExecutor]]
 
     with pool_executor() as executor:
         for chunk in chunks(pii_data, chunk_size):
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -187,7 +187,7 @@ A corresponding hashing schema can be generated as well::
             "weight": 0
           }
         },
-        {s
+        {
             "minLength": 3
           },
           "hashing": {
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -188,6 +188,11 @@ A corresponding hashing schema can be generated as well::
           }
         },
         {
+          "identifier": "NAME freetext",
+          "format": {
+            "type": "string",
+            "encoding": "utf-8",
+            "case": "mixed",
             "minLength": 3
           },
           "hashing": {
@@ -221,6 +226,7 @@ A corresponding hashing schema can be generated as well::
         }
       ]
     }
+
 
 
 Benchmark
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,9 @@ nosetests.xml
 
 # Operating system stuff
 .DS_Store
+
+# Hypothesis
+.hypothesis/
+
+# Docs build
+docs/_build/
--- a/docs/doc-requirements.txt
+++ b/docs/doc-requirements.txt
@@ -1,7 +1,7 @@
 anonlink>=0.12
-ipython==7.5.0
+ipython==7.12.0
 jupyter
-matplotlib==3.1.1
+matplotlib==3.1.3
 nbsphinx>=0.3
 numpy
 pandas
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 bashplotlib==0.6.5
-bitarray-hardbyte==1.1.0
+bitarray-hardbyte==1.2.2
 click==7.0
 cryptography==2.8
 enum34==1.1.6; python_version < '3.4'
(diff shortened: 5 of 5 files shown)


doc-E-brown:
Build failures are mostly due to a missing codecov token in the build pipeline as well as one pytest timeout.

clkhash/cli.py
```diff
@@ -141,8 +141,9 @@ def cli(verbose):
 @click.option('--no-header', default=False, is_flag=True, help="Don't skip the first row")
 @click.option('--check-header', default=True, type=bool, help="If true, check the header against the schema")
 @click.option('--validate', default=True, type=bool, help="If true, validate the entries against the schema")
+@click.option('--multiprocessing', default=True, type=bool, help="If true, use multiprocessing to generate the keys from the secret")
```
hardbyte:
```suggestion
@click.option('--multiprocessing', default=True, type=bool, help="If true, use multiprocessing to generate encodings")
```

clkhash/clk.py
```diff
@@ -75,6 +76,11 @@ def generate_clk_from_csv(input_f,  # type: TextIO
             header but it should not be checked against the schema.
         :param bool progress_bar: Set to `False` to disable the progress
             bar.
+        :param bool use_multiprocessing: Set to `False` to use threading to
+            generate keys for each identifier from the secret.  By default
+            multiprocessing is used to generate the keys.  Set this flag to
+            `False` if executing on a system that is not capable of spawning
+            subprocesses such as AWS Lambda functions.
```
hardbyte:
```suggestion
        :param bool use_multiprocessing: Set to `False` to use threading instead
            of multiprocessing to generate encodings for each record. Set this flag 
            to `False` if executing on a system that is not capable of spawning
            subprocesses such as AWS Lambda functions.
```

clkhash/clk.py
```diff
@@ -154,7 +165,10 @@ def generate_clks(pii_data,  # type: Sequence[Sequence[str]]
     futures = []
 
     # Compute Bloom filter from the chunks and then serialise it
-    with concurrent.futures.ProcessPoolExecutor() as executor:
+    pool_executor = ProcessPoolExecutor if use_multiprocessing else \
+        ThreadPoolExecutor # type: Union[Type[ProcessPoolExecutor], Type[ThreadPoolExecutor]]
+
+    with pool_executor() as executor:
```
hardbyte: Nice

docs/cli.rst
```diff
@@ -187,12 +187,7 @@ A corresponding hashing schema can be generated as well::
             "weight": 0
           }
         },
-        {
-          "identifier": "NAME freetext",
-          "format": {
-            "type": "string",
-            "encoding": "utf-8",
-            "case": "mixed",
+        {s
```
hardbyte: Assume this was a mistake?

docs/cli.rst
```diff
@@ -250,6 +243,16 @@ As a rule of thumb a single modern core will hash around 1M entities in about 20
...
 as a basic sanity check - ensure the CLK's popcount is not around 0 or 1024.
 
+By default the benchmark uses multiprocessing to generate the clks, the performance difference by switching
+to multi-threading mode can also be determined by passing the command line option ``--multiprocessing 0``.
+
+    python -m clkhash.cli benchmark --multiprocessing 0
+    generating CLKs: 100%                 10.0K/10.0K [00:05<00:00, 1.90Kclk/s, mean=406, std=20.4]
+     10000 hashes in 5.285763 seconds. 1.89 KH/s
+
+The performance reduction in using multithreading mode is significant however it does allo clks to be
```
hardbyte:
```suggestion
The performance reduction in using multithreading mode is significant however it does allow clks to be 
```

docs/cli.rst
doc-E-brown: Yes, not sure how this ended up here.  I will remove

clkhash/clk.py
doc-E-brown: Thanks!

docs/cli.rst
hardbyte: The deletion as well thanks

doc-E-brown: Sorry! Corrected

hardbyte:
We should fix the codecov token issue - but I'll just push your changes into a different branch to avoid blocking the merge.

doc-E-brown:
Awesome, thanks!

hardbyte:
I've merged a fix to the codecov token bug - if you want to update we *should* see the CI pass.

hardbyte:
Also please note @wilko77's comments in https://github.com/data61/clkhash/pull/343#pullrequestreview-367699607

hardbyte:
> > It's great that you made clkhash work in lambda functions.
> > A typical clkhash user might not know the difference between multi-processing and multi-threading. So this all might be a bit confusing. Do we really need to expose that switch in the CLI? Do you call clkhash in the AWS lambda function via the CLI?
> > I think we should provide more info in the CLI and docs, if we want to keep that switch there.
> 
> We are directly using the generate_clks function in the lambda function not the CLI. I agree about more info within the docs and I will update the PR accordingly. The reason for adding the switch in the CLI was to allow quick testing of the performance drop in using threads as opposed to multiprocesses to see if this is acceptable. I am happy to add this information to the docs as well or just remove the switch altogether.

Cool, lets just remove the CLI switch then good to merge.

clkhash/clk.py
```diff
@@ -75,6 +76,10 @@ def generate_clk_from_csv(input_f,  # type: TextIO
             header but it should not be checked against the schema.
         :param bool progress_bar: Set to `False` to disable the progress
             bar.
+        :param bool use_multiprocessing: Set to `False` to use threading instead
+            of multiprocessing to generate encodings for each record. Set this flag 
+            to `False` if executing on a system that is not capable of spawning
+            subprocesses such as AWS Lambda functions.
```
hardbyte: Picking up the conversation from https://github.com/data61/clkhash/pull/343#discussion_r386796625
@wilko77 @doc-E-brown 
Perhaps a cleaner approach than introducing another rarely used optional variable is to replace `use_multiprocessing` with an executor argument - expect an optional  [`Executor` instance](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor) (with the Process Pool Executor being the default).

doc-E-brown: Sounds like a good approach to me, I'll update soon.

doc-E-brown: Just to touch base, I haven't forgotten about this, just slammed at work.  Will update first chance I get.

hardbyte: Fair enough! The whole world is a bit crazy right now 😷 
Thanks for the update and just let us know if you'd prefer us to implement the suggested changes.

PR #342, opened 2020-02-28
Fix schema v3

### LINKED ISSUE #341: schema validation of 'ignored' can fail
We have two ways of defining a feature to be ignored in the schema.
1) we just add `"ignored": true` to the top level of the feature def.
e.g.:
```
{
      "identifier": "firstName",
      "format": { ... },
      "hashing": { ... },
      "ignored": true
},
```
2) as a shortcut, we can also just have "identifier" and "ignored"
e.g.:
```
 {
      "identifier": "id",
      "ignored": true
},
```

The way the schema def is written, the validator cannot disambiguate between the definition for 1) and 2) if you provide the example for 1).
You get an error saying: `... is valid under each of {'$ref': '#/definitions/ignoreFeature'}, {'$ref': '#/definitions/featureConfig'}`

wilko77:
fixes #341

the schema validator couldn't decide which spec to apply if `"ignored": true` was added to a full feature config. More detail in #341

### DIFF
--- a/clkhash/schemas/v3.json
+++ b/clkhash/schemas/v3.json
@@ -72,6 +72,7 @@
       "properties": {
         "identifier": {"type": "string", "description": "semantic meaning of identifier"},
         "description": {"type": "string"},
+        "ignored": {"type": "boolean"},
         "format": {
           "required": ["type"],
           "oneOf": [
--- a/clkhash/schemas/v3.json
+++ b/clkhash/schemas/v3.json
@@ -57,6 +57,7 @@
     "ignoreFeature": {
       "type": "object",
       "required": ["identifier", "ignored"],
+      "additionalProperties": false,
       "properties": {
         "identifier": {"type": "string", "description":"semantic meaning of identifier"},
         "ignored":  {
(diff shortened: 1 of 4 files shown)


tests/test_schema.py
```diff
@@ -138,6 +138,11 @@ def test_validation_of_illdefined_not_ignored_feature(self):
         exception = contextmanager.exception
         self.assertIsInstance(exception, SchemaError)
 
+    def test_ignore_definitions(self):
+        # there are several valid ways of specifying the ignored property.
+        with open(_test_data_file_path('ignorant-schema-v3.json'), 'r') as f:
+            schema.from_json_file(f, validate=True)
```
hardbyte: As well as checking that the schema is valid, also confirm that the expected features are indeed ignored.

clkhash/schemas/v3.json
```diff
@@ -57,6 +57,7 @@
     "ignoreFeature": {
       "type": "object",
       "required": ["identifier", "ignored"],
+      "additionalProperties": false,
```
hardbyte: Won't the validator now decide a fully defined feature that includes a `ignored=true` will be an instance of `featureConfig` rather than `ignoreFeature`? That doesn't seem ideal?

wilko77: I want the three different types of definitions as in the `testdata/ignorant-schema-v3.json` file to be possible but not allow
```
{
  "identifier": "id",
  "ignored": false
}
```
I made this more explicit now in the schema spec.

PR #345, opened 2020-03-02
Use codecov.io's bash uploader

hardbyte:
The bash uploader from codecov supports Azure without needing to configure secret tokens. This should resolve the issue seen in #340 where an external PR didn't pass CI.

### DIFF
--- a/.azurePipeline/wholeBuild.yml
+++ b/.azurePipeline/wholeBuild.yml
@@ -87,8 +87,8 @@ jobs:
       codeCoverageTool: Cobertura
       summaryFileLocation: 'coverage.xml'
       failIfCoverageEmpty: true
-  - script: |
-      bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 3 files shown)


.azurePipeline/wholeBuild.yml
```diff
@@ -85,21 +85,12 @@ jobs:
...
   - bash: |
-        report_name="$(operatingSystem)_$(architecture)_$(pythonVersion)"
-        if [ $(fullTests) == "True" ]
-        then
-          report_name=${report_name}_full_tests
-        fi
-        python -m codecov --token $(CODECOV_TOKEN) \
-          --file coverageReport.xml \
-          -n ${report_name}
-    displayName: 'Send coverage to codecov'
```
JoyceXinyueWang: haha with pytestazure, we don't need this anymore. Blocklib and anonlink-client already omitted this section!

PR #348, opened 2020-03-30
Remove cli and rest_client from clkhash

hardbyte:


clkhash/backports.py
```diff
@@ -39,11 +39,15 @@ def _int_from_bytes(bytes, byteorder, signed=False):
         hex_str = codecs.encode(bytes, 'hex')  # type: ignore
         return int(hex_str, 16)
 
-
+    try:
+        from typing import SupportsBytes
+        bytes_arg_type = Union[Iterable[int], SupportsBytes]
+    except ImportError:
+        bytes_arg_type = Iterable[int]
```
JoyceXinyueWang: Can you explain what this code is for? I am a bit confused here.

docs/development.rst
```diff
@@ -23,13 +23,12 @@ Now run the unit tests and print out code coverage with `py.test`::
...
 
-Note several tests will be skipped by default. To enable the command
-line tests set the  `INCLUDE_CLI` environment variable. To enable
+Note several tests will be skipped by default. To enable
 the tests which interact with an entity service set the
 `TEST_ENTITY_SERVICE` environment variable to the target service's
 address::
 
-    $ TEST_ENTITY_SERVICE= INCLUDE_CLI= python -m pytest --cov=clkhash
+    $ TEST_ENTITY_SERVICE= python -m pytest --cov=clkhash
```
JoyceXinyueWang: Shouldn't we remove TEST_ENTITY_SERICE entirely?

hardbyte: Yup good catch, I'll search the whole project for this and remove.

clkhash/backports.py
hardbyte: It was a (failed) attempt to make new and old versions of both Python and mypy all co-exist. I'll remove this kludge and update mypy so this should just revert back.

azurePipeline.yml
```diff
@@ -50,13 +50,11 @@ stages:
       pythonVersions: ['3.8', '2.7']
       operatingSystems: ['ubuntu-18.04']
       architectures: ['x64']
-      fullTests: True
   - template: .azurePipeline/wholeBuild.yml  # Template reference
     parameters:
       pythonVersions: ['3.7']
       operatingSystems: ['ubuntu-18.04', 'macos-10.14', 'vs2017-win2016']
       architectures: ['x64']
```
wilko77: you could sort those versions into the entries below. Would be a bit neater.

docs/tutorials.rst
```diff
@@ -1,22 +1,29 @@
...
+
+The tutorial `tutorial_api.ipynb` shows an example linkage workflow.
 
 With linkage schema version 3.0 *clkhash* introduced different comparison techniques for feature values.
 They are described in the tutorial `tutorial_comparisons.ipynb`.
 
+The notebooks can run online using binder.
+
+.. image:: https://mybinder.org/badge_logo.svg
+  :target: https://mybinder.org/v2/gh/data61/clkhash/master?filepath=docs
```
wilko77: I would put this section under the `running the tutorials` heading.

setup.py
```diff
@@ -7,16 +7,13 @@
 requirements = [
         "bashplotlib>=0.6.5",
```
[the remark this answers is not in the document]
hardbyte: We used it in the "public" api `clkhash.describe.plot` - but I'll change that to simply return the counts.

PR #360, opened 2020-06-16
Goodbye python2

wilko77:
we remove python2 support. It's dead. 

- Removed all the backports. 
- Removed all the python2 testing from the CI
- Removed dependencies only used with python2
- modernized the typing annotations


clkhash/bloomfilter.py
```diff
@@ -26,13 +24,12 @@
...
-def double_hash_encode_ngrams(ngrams,   # type: Iterable[str]
-                              keys,     # type: Sequence[bytes]
-                              ks,       # type: Sequence[int]
-                              l,        # type: int
-                              encoding  # type: str
-                              ):
-    # type: (...) -> bitarray
+def double_hash_encode_ngrams(ngrams: Iterable[str],
+                              keys: Sequence[bytes],
+                              ks: Sequence[int],
```
JoyceXinyueWang, line 29: Nice! I prefer this than the comment style type annotation

clkhash/clk.py
```diff
@@ -85,7 +82,7 @@ def generate_clk_from_csv(input_f,  # type: TextIO
     log.info("Hashing data")
 
     # Read from CSV file
-    reader = unicode_reader(input_f)
+    reader = csv.reader(input_f)
```
JoyceXinyueWang, line 85: What is the reason for using `csv` over `unicode_reader`?

wilko77: unicode_reader was a workaround for python2 to be able to read unicode in csv files. In python3 the csv.reader can do that natively.

clkhash/bloomfilter.py
wilko77: me too. Unfortunately, in python 3.5 you cannot yet specify types of variables like this:
`n:int = 10` :(
We still need to use the comment style in those cases

Issue #321, opened 2019-11-18
Testing notebooks on Windows machine in CI


gusmith:
Currently (with master branch), the notebooks cannot install anonlink (which requires a compiler).
When anonlink v0.13 will be released, the issue should be fixed as it will include a windows wheel.

hardbyte:
FYI should be able to test this using the prerelease from PyPi `anonlink==0.13.0.dev0`

wilko77:
anonlink is fine. Now something else fails. See https://dev.azure.com/data61/Anonlink/_build/results?buildId=2506&view=logs&j=556e8606-17d2-50a8-f13f-ecd65b797e53

wilko77:
closed via  #377

PR #377, opened 2020-08-07
test notebooks on windows also

wilko77:
addresses issue #321

### DIFF
--- a/.azurePipeline/wholeBuild.yml
+++ b/.azurePipeline/wholeBuild.yml
@@ -51,7 +51,7 @@ jobs:
       pytest --nbval docs -x --sanitize-with docs/tutorial_sanitize.cfg
     displayName: 'Test all notebooks'
     timeoutInMinutes: 5
-    condition: and(ne(variables['pythonVersion'], '3.5'), ne(variables['operatingSystem'], 'vs2017-win2016'), ne(variables['pythonVersion'], 'pypy3'))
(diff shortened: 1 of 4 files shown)


.azurePipeline/wholeBuild.yml
```diff
@@ -46,12 +46,12 @@ jobs:
...
 # http://packages.pypy.org/##pandas
   - script: |
-      pip install -e .
+      python -m pip install -U -e .
       python -m pip install -U -r docs/doc-requirements.txt
       pytest --nbval docs -x --sanitize-with docs/tutorial_sanitize.cfg
     displayName: 'Test all notebooks'
     timeoutInMinutes: 5
-    condition: and(ne(variables['pythonVersion'], '3.5'), ne(variables['operatingSystem'], 'vs2017-win2016'), ne(variables['pythonVersion'], 'pypy3'))
+    condition: and(ne(variables['pythonVersion'], '3.5'), ne(variables['pythonVersion'], 'pypy3'))
```
JoyceXinyueWang, line 54: This line of change looks like removing test on windows?

wilko77: it's a bit hard to read, but the `ne()` part means "not equals". Thus, previously, we ran the notebooks if 'operatingSystem' was not equal 'vs2017-win2016'.

JoyceXinyueWang (approved):

PR #383, opened 2020-08-28
Fix strategy definition

wilko77:
the way the strategy was defined in the schema, one could specify
```"strategy": {"bitsPerToken": "illegalValue"}```
and the validation would not complain.
This PR fixes the definition of strategy in the schema.


JoyceXinyueWang (approved):
Good catch!

PR #385, opened 2020-09-10
fix entry for numeric comparison

wilko77:
one parameter was missing and another one was misnamed.


JoyceXinyueWang (approved):
Look good to me!

PR #391, opened 2020-10-06
fix deprecation warning, reduced hypothesis test cases range to somet…

wilko77:
…hing more sensible.

Testing with such huge ranges for threshold distance and candidates leads to some floating point weirdness. I think we can just test for numbers up to +/-1e20. That should cover most realistic use cases.

### DIFF
--- a/tests/test_comparators.py
+++ b/tests/test_comparators.py
@@ -16,7 +16,7 @@ from clkhash.comparators import NgramComparison, NonComparison, ExactComparison,
 #######
 
 
-@fixture(params=itertools.product((1, 2, 3), (True, False)))
+@fixture(params=itertools.product((1, 2, 3), (True, False)), scope='module')
 def ngram_comparator(request):
     return comparators.NgramComparison(request.param[0], request.param[1])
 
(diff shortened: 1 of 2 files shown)


tests/test_comparators.py
```diff
@@ -118,10 +118,10 @@ def test_numeric_mix_int_and_floats():
 
 # we restrict the exponents of the floats such that exponent of number + precision < 308. Otherwise we might get
 # infinity during the tokenization process.
-@given(thresh_dist=floats(min_value=0.0, allow_nan=False, allow_infinity=False, max_value=1e287, exclude_min=True),
+@given(thresh_dist=floats(min_value=0.0, allow_nan=False, allow_infinity=False, max_value=1e20, exclude_min=True),
        resolution=integers(min_value=1, max_value=256),
        precision=integers(min_value=0, max_value=20),
-       candidate=floats(allow_nan=False, allow_infinity=False, max_value=1e288, min_value=-1e287))
+       candidate=floats(allow_nan=False, allow_infinity=False, max_value=1e20, min_value=-1e20))
```
JoyceXinyueWang, line 124: from 1e287 to 1e20, it looks like a huge decrease. Are we only testing integer here?

tests/test_comparators.py
```diff
@@ -135,19 +135,24 @@ def test_numeric_properties(thresh_dist, resolution, precision, candidate):
...
+@given(thresh_dist=floats(allow_infinity=False, allow_nan=False, min_value=0.0, max_value=1e20, exclude_min=True),
        resolution=integers(min_value=1, max_value=256),
        precision=integers(min_value=0, max_value=20),
-       candidate=floats(allow_infinity=False, allow_nan=False, max_value=1e287, min_value=-1e287))
+       candidate=floats(allow_infinity=False, allow_nan=False, max_value=1e20, min_value=-1e20))
 def test_numeric_overlaps_around_threshdistance(thresh_dist, resolution, precision, candidate):
     assume(int(round(thresh_dist * pow(10, precision))) > 0)
     comp = NumericComparison(threshold_distance=thresh_dist, resolution=resolution, fractional_precision=precision)
     other = candidate + thresh_dist
+    if other - candidate < thresh_dist:  # test for floating point errors
```
JoyceXinyueWang, line 146: but you just defined other = candidate + thresh_dist. This would always be false?

JoyceXinyueWang (commented):
Could you answer my questions before I approve it?

tests/test_comparators.py
wilko77: not with floating point numbers.
E.g.:
```
candidate = 1.9559898792067546e+17
thresh_dist = 22204.460492503134
other = candidate + thresh_dist
actual_dist = other - candidate
actual_dist
Out[6]: 22208.0
```

tests/test_comparators.py
wilko77: definitely still floating point numbers. Just smaller. Can you think of any linkage features that are measured in floats larger than 10 ** 20?

JoyceXinyueWang (approved):

PR #406, opened 2020-12-09
remove testing for python 3.5

wilko77:
support for 3.5 ended in September. 
In the meantime, they released 3.9. We better test that then...


joyceyuu (approved):
Look good to me!

PR #423, opened 2021-01-25
Feature allow thread pool executor

### LINKED ISSUE #339: Use of ProcessPoolExecutor prevents use in Lambda functions

hardbyte:
As discussed in https://github.com/data61/clkhash/pull/340, this closes #339.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+
+
+## 0.16.1
+
+`generate_clk_from_csv` and `generate_clks` now accept an optional Executor instance.
+This means systems that can't create sub-processes such as celery workers and AWS lambda
+jobs can now use clkhash.
+
 ## 0.16.0
 
 Removes `rest_client` and `cli` modules. This functionality has
--- a/clkhash/benchmark.py
+++ b/clkhash/benchmark.py
@@ -1,17 +1,20 @@
 import os
 import tempfile
 # noinspection PyProtectedMember
+from concurrent.futures.thread import ThreadPoolExecutor
 from timeit import default_timer as timer
 
 from clkhash.clk import generate_clk_from_csv
 from clkhash.randomnames import NameList
 
 
-def compute_hash_speed(num: int, quiet: bool = False) -> float:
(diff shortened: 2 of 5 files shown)


wilko77 (changes requested):
I don't think `ThreadPoolExecutor` is a good choice here. Because of the GIL, the standard Python interpreter will only ever run one thread at any given time. Thus spreading the work over several threads will never be faster than just doing all the work in the main thread. In fact, it is quite a bit slower, especially with the default arguments for ThreadPoolExecutor (before py38 it defaults to 5 * numProcessors threads). 

If I understand the problem correctly, then AWS lambdas and celery workers don't allow multiprocessing. And they are designed to execute one thread of code.

Wouldn't it be simpler to just add a `max_workers` argument to `generate_clks`. In case of 1, just run everything in the main loop, otherwise pass it on to ProcessPoolExecutor. 

(This would also allow us to tame clkhash if needed. Currently it just grabs all available cores.)

hardbyte:
Could be worth testing, I had assumed that the bulk of the work would be in the low level crypto calls (HMAC etc) and that they would be releasing the GIL which would allow threads to still be useful.

`max_workers` argument would be a nicer api though so happy to implement that

hardbyte:
You were correct, it is faster using no threads.

Informal numbers:
- Passing max_workers=4 to ProcessPool ~ 10kCLK/s
- passing max_workers=8 to ProcessPool ~12kCLK/s
- Using default ThreadPool ~1.9 kCLK/s
- Main loop (no executor) ~4kCLK/s
Replacement PR here - https://github.com/data61/clkhash/pull/424

PR #424, opened 2021-01-28
Feature max workers

hardbyte:
Replaces https://github.com/data61/clkhash/compare/feature-max-workers?expand=1

### DIFF
--- a/clkhash/benchmark.py
+++ b/clkhash/benchmark.py
@@ -1,17 +1,20 @@
 import os
 import tempfile
 # noinspection PyProtectedMember
+from concurrent.futures.thread import ThreadPoolExecutor
 from timeit import default_timer as timer
 
 from clkhash.clk import generate_clk_from_csv
 from clkhash.randomnames import NameList
 
 
-def compute_hash_speed(num: int, quiet: bool = False) -> float:
+def compute_hash_speed(num: int, quiet: bool = False, use_multiprocessing=True) -> float:
     """ Hash time.
     """
     namelist = NameList(num)
-
+    executor = None
+    if not use_multiprocessing:
+        executor = ThreadPoolExecutor()
     os_fd, tmpfile_name = tempfile.mkstemp(text=True)
 
     schema = NameList.SCHEMA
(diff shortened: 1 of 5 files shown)


clkhash/benchmark.py
```diff
@@ -40,3 +43,6 @@ def compute_hash_speed(num: int, quiet: bool = False) -> float:
 if __name__ == '__main__':
     for n in [100, 1000, 10000, 50000, 100000]:
         compute_hash_speed(n, quiet=n <= 10000)
+
+    print("Without multiprocessing")
+    compute_hash_speed(10000, quiet=n <= 10000, use_multiprocessing=False)
```
wilko77, line 48: I'm not sure if you intended to quieten the output here. Wouldn't it be more useful if the user can see the timings?

clkhash/clk.py
```diff
@@ -6,6 +6,7 @@
 import csv
 import logging
 import time
+from concurrent.futures._base import Executor
```
wilko77, line 9: is that still used somewhere?

setup.py
```diff
@@ -40,6 +40,7 @@
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
```
wilko77, line 43: nice :)

clkhash/clk.py
```diff
@@ -73,6 +75,10 @@ def generate_clk_from_csv(input_f: TextIO,
             header but it should not be checked against the schema.
         :param bool progress_bar: Set to `False` to disable the progress
             bar.
+        :param int max_workers: Passed to ProcessPoolWorker except for the
```
wilko77, line 78: `ProcessPoolExecutor`, not worker.

wilko77 (approved):
Great, thanks. 
Looks good. Just a few minor things in the comments.

clkhash/benchmark.py
hardbyte: The timings are printed because `n` is always `100_000` at this point, but yeah shouldn't have been copied. I'll clean that up.

setup.py
hardbyte: Was tempted to add 3.10 too :-)

PR #561, opened 2022-03-10
GitHub action for testing and publishing

wilko77:
Moving away from the Azure pipeline.
I recreated most functionality:

- a workflow for typechecking
- a workflow for unittesting
  - we run the normal pytest across a matrix of os and python versions. I used pypy-3.8, as pypy3 is deprecated and point to an ancient pypy3.6.
  - we run the notebook tests only on the latest os'es with python 3.10, as these tests take quite a bit of time
- a workflow to build the distribution and publish
  - there are two triggers, PR's into main and release published. We build on both, but only publish to PyPi on release.


.github/workflows/build_publish.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install build
+      - name: Build package
+        run: python -m build
```
hardbyte, line 23: Oh haven't come across `build` nice find.

.github/workflows/unittests.yml
```diff
@@ -0,0 +1,95 @@
...
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  unittests:
+    name: Unittests ${{ matrix.python }} ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-18.04, ubuntu-20.04]
```
hardbyte, line 18: Is there much point keeping `ubuntu-18.04`?

.github/workflows/unittests.yml
```diff
@@ -0,0 +1,95 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python }}
+
+      - name: Get full Python version
+        id: full-python-version
+        shell: bash
+        run: echo ::set-output name=version::$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))")
```
hardbyte, line 80:
```suggestion
```

.github/workflows/unittests.yml
```diff
@@ -0,0 +1,95 @@
...
```
hardbyte, line 36: I don't think this step is required?
```suggestion

```

hardbyte (approved):

.github/workflows/unittests.yml
wilko77: it's still supported until 2023. But then, so are many older mac and windows versions.
I think we're gonna be fine only using `ubuntu-latest`.

PR #673, opened 2023-05-09
read clks in a streaming fashion

wilko77:
Not sure about the API change. Maybe introduce an extra function?

However, this way of doing things is:
- faster: I measured up to 20% for large files
- less memory demand: I measured about 33% memory consumption compared to std clkhash.


clkhash/clk.py
```diff
@@ -22,10 +25,11 @@
 log = logging.getLogger('clkhash.clk')
 
 
-
 def hash_chunk(chunk_pii_data: Sequence[Sequence[str]],
                keys: Sequence[Sequence[bytes]],
-               schema: Schema
+               schema: Schema,
+               validate_data: bool,
+               row_index_offset: int
```
hardbyte, line 32: I like the addition of `row_index_offset` - but it took me a beat to realize it is just for reporting the location of validation errors and otherwise doesn't effect the results. I'll push a docstring.

wilko77:
superseded by #675

